Issue #5, opened 2022-04-01
Add docker-compose files to build locally


ioggstream:
## I expect

docker compose files to build / test locally

PR #6, opened 2022-04-01
Fix: #5. Build with docker.

ioggstream:
## This PR

builds with docker locally

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+build
+dist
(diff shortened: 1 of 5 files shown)


netfl0 (approved):
looks great, thank you for the cleanup as well :)

ioggstream:
@netfl0  you welcome! Give it a spin, since I just ran it on my laptop :) You can probably create a github-ci out of it using a gh-action that just runs docker-compose up :)

PR #8, opened 2022-04-07
Add SAST + DAST.

ioggstream:
## This PR

- adds SAST + DAST

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -13767,14 +13767,15 @@ Data Execution Prevention (DEP) does exactly that, by substantially reducing the
 ###  http://d3fend.mitre.org/ontologies/d3fend.owl#StaticApplicationSecurityTesting
 :StaticApplicationSecurityTesting rdf:type owl:NamedIndividual ,
                               :ApplicationHardening ;
-                     :kb-reference :Reference-StaticApplicationSecurityTesting-TODO ;
+                     :kb-reference :Reference-StaticApplicationSecurityTesting ;
                      :d3fend-id "D3-SAST" .
 
 ###  http://d3fend.mitre.org/ontologies/d3fend.owl#DynamicApplicationSecurityTesting
 :DynamicApplicationSecurityTesting rdf:type owl:NamedIndividual ,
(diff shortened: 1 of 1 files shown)


netfl0:
Do you have proposed definitions and digital artifact mappings for these?

netfl0:
I am back and was able to review the pull request, thank you for your patience.

In the ontology we make a distinctions between d3f:Application, d3f:ExecutableFile, and (soon to exist) d3f:SourceCode. We are going to be adding a Source Code Hardening as a sibling to Application Hardening.

I am thinking we should enumerate the essential categories of tests the DAST and SAST tools run before we develop the higher level category, we then can model what the tests are actually testing. Input Length (Limiting | Validation | Verification) might be one test case for example. Then, we are able to model the relevant digital artifacts. I don't think we will be able to model the digital artifacts if we just use DAST and SAST as techniques.

The same argument could be made for our current File Analysis technique subclass: DynamicAnalysis, so there may be an opportunity to improve that as well.

I am thinking we should enumerate the essential categories of tests the DAST and SAST tools run before we develop the higher level category, we then can model what the tests are actually testing. Then, we are able to model the relevant digital artifacts. I don't think we will be able to model the digital artifacts if we model DAST and SAST as DefensiveTechniques.

ioggstream:
@netfl0 

> Source Code Hardening as a sibling to Application Hardening

SourceCodeHardening subClassOf ApplicationHardening ?

> enumerate the essential categories of tests the DAST and SAST tools run

since some of those tests are actually implemented by developers (e.g. DAST can be actually anything the developers implements) probably adding "unstable" resources can be a good way to start.

> .. SAST ...

SAST tests can be easier to identify, eg. linting, secrets disclosure, dependency management, ...

netfl0:
Marking this as OBE, some of the concepts like containers are now covered. More detailed testing taxonomies will be required in the future. Please also see the related new Source Code Hardening taxonomy

ioggstream:
+[OBE](https://www.merriam-webster.com/dictionary/overtaken%20by%20events#:~:text=%3A%20forced%20to%20be%20changed%20because,and%20had%20to%20be%20changed) ;)

Issue #26, opened 2022-05-04
Inbound Traffic Filtering CM Incorrectly Maps to 12 Exfiltration ATT&CK Techniques


OldGoat-Biker:
Because the artifact association is just at the network traffic level, it makes no differentiation of direction of traffic and ATT&CK techniques that only operate on one flow direction. Inbound Traffic Filtering does nothing to counter the 12 Exfil threats with which it is currently associated. Need to scrutinize this (like others) threat to artifact to CM association more closely. Probably needs a more specific  relationship operator.

PR #27, opened 2022-05-04
refs #26, refactored ITF and OTF under new technique D3-NTF, added ne…

netfl0:
Please review.


hack-sentinel:
Minor bug on branch. Have edited and committed locally on branch, ready to push for eventual merge.  

:boom: But cannot push to branch, fails with 403.  Can't verify have any permissions on project repo, and "Insights" doesn't show this to me so maybe I need to be added as code committer-worthy?  [git let me update with pull, so at least I could read latest on the branch and update tooling and edit locally.]

Issue #1, opened 2022-02-17
[Ontology] Update D3-SPP Artifact Mappings.


netfl0:
Contributor:
Why is Strong Password Policy not mapped to Password Cracking in the Knowledge Graph?

hack-sentinel:
Reviewed and modified Credential Access Technique, Brute Force, and Brute Force subclasses to indicate more precisely the main digital artifacts being accessed (here tested by some means of trial and error.)  Result with query from D3-SPP now results in new matches the the password-specific classes under Brute Force.

![image](https://user-images.githubusercontent.com/101665567/167186614-e81a426e-263a-4f00-a617-a61f735c563c.png)

PR #30, opened 2022-05-06
Reviewed and modified Credential Access Technique, Brute Force, and Brute Force subclasses

hack-sentinel:
Edits will  indicate more precisely the main digital artifacts being accessed (here the passwords are tested by some means of trial and error.)  Result with defense-to-offense query from D3-SPP now results in new matches the the password-specific classes under Brute Force.


netfl0:
I believe this relates to #1 .

Issue #37, opened 2022-05-12
ContainerImageAnalysis < FileAnalysis, ApplicationHardening


ioggstream:
## I Expect

ContainerImageAnalysis from NSA Kubernetes Security

## Notes

This can be both:

- Harden, when applied to the Software Delivery Lifecycle
- Detect, when applied to existing application deployments 
- Detect: when applied to existing platform deployments

The boundary b/w Harden and Detect here is blurred to me

netfl0:
The biggest distinction for Detect (vs Harden) is that it is detection **Adversary Activity**, not things likes vulnerabilities or weaknesses.

> Definition
The detect tactic is used to identify adversary access to or unauthorized activity on computer networks.

https://d3fend.mitre.org/tactic/d3f:Detect

ioggstream:
Thus we need a "preventive FileAnalysis" measure.

PR #38, opened 2022-05-12
Fix: #37. Add ContainerImageAnalysis kb.

ioggstream:
## This PR

- adds ContainerImage Analysis

Fixes: #37


netfl0:
In the kb-article it was noted to scan for weak configurations (we do have D3-ACH for fixing the problems discovered) and vulnerable libraries (D3-SU is the fix here). Those are distinct countermeasures  we should probably break out. Given that applications are inside the containers and other package formats, we may need to break this out to make sure its obvious which one to select. We are working on vulnerability scanning / inventory, tagging @hack-sentinel  since he is currently working in this space.

ioggstream:
To address this, we should define whether ContainerImages fall under PlatformHardening or ApplicationHardening.

IMHO:

- if we deploy a containerized application, then the whole container is probably the deliverable, and the container image should be treated like software libraries;
- if we deploy a containerized system functionality, then  D3-SU is probably corret.

Issue #40, opened 2022-05-17
Which artifact for an Executable interpreter


ioggstream:
## Question

Which artifact maps to an executable interpreter e.g. Java Virtual Machine or Python?

1. :ExecutableFile mentions "in a more general sense, a file containing instructions (such as bytecode) for a software interpreter may also be considered executable" :thinking: 
2. :ExecutableScript mentions that a script "is different from an ExecutableFile :+1: 

afaik

1.  java do not fall into this model since it executes bytecode (assembled by somebody else)
2.  python currently converts it into bytecode before executing, but this is just nitpicking

## Note

This is useful to define the artifacts associated to python and java libraries, and to answer to the following questions:

- interpreted and bytecode libraries should be associated to different artifacts?
- since dependencies might contain shared libs, could a more general concept of "application dependency" be useful in this case? 
- could the above artifact (ApplicationDependency) apply to source code dependencies that are statically compiled at build time together with the application code (e.g. golang?)

netfl0:
> 



> 



> ## Question
> 
> Which artifact maps to an executable interpreter e.g. Java Virtual Machine or Python?
> 
>     1. :ExecutableFile mentions "in a more general sense, a file containing instructions (such as bytecode) for a software interpreter may also be considered executable" 🤔
> 
>     2. :ExecutableScript mentions that a script "is different from an ExecutableFile 👍
> 
> 
> afaik
> 
>     1. java do not fall into this model since it executes bytecode (assembled by somebody else)
> 
>     2. python currently converts it into bytecode before executing, but this is just nitpicking
> 
> 
> ## Note
> 
> This is useful to define the artifacts associated to python and java libraries, and to answer to the following questions:
> 
>     * interpreted and bytecode libraries should be associated to different artifacts?

Thus far, we have not needed the distinction, but with the contributions you're working on, it is clear we need the additional types. Agree?

> 
>     * since dependencies might contain shared libs, could a more general concept of "application dependency" be useful in this case?
Yes, I was thinking something like Software Library, which would contain Software Library Files( we already have  -> Shared Library File).  Software Libraries might also contain configuration files as well. These would be similar to but distinct from Application Configuration Files.

> 
>     * could the above artifact (ApplicationDependency) apply to source code dependencies that are statically compiled at build time together with the application code (e.g. golang?)

[...]

PR #45, opened 2022-05-23
Add SoftwareLibrary and dependencies. See #40.

ioggstream:
## This PR

SoftwareLibraryFile < File:
  contains: Software
  may-contain: ExecutableBinary, SourceCode

SoftwareLibrary < Software
  contains: SoftwareLibraryFile

JavaArchive < ExecutableBinary
PythonPackage < ExecutableBinary

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -5013,6 +5013,12 @@ IPC can occur within a single computer or between multiple computers remotely th
     :definition "The isolate tactic creates logical or physical barriers in a system which reduces opportunities for adversaries to create further accesses." ;
     :display-order 2 .
 
+:JavaArchive a owl:Class ;
+    rdfs:label "Java Archive" ;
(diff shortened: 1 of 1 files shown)


ioggstream:
@hack-sentinel I'd like to integrate  SBOM / dependency management in d3fend. 

- Does SoftwareInventory apply to software dependencies too?
- how should I modify this PR in order to be merged?

hack-sentinel (approved):
Looks good (includes backref to #40.)  At some point it may be useful distinguish bytecode from direct executables, but we'd also have to tackle virtual machine concepts (Java and VMWare different senses and layerings) and in some contexts users might need means to explicitly distinguish virtual from physical.

Immediately before or after merging this to develop, we need to translate rdfs:comment property assertions to d3f:definition assertions before pushing to release.

hack-sentinel:
Merged content to 0.12.0-BETA-1 as well.

PR #68, opened 2022-09-16
Feature/netfl0 63 add 800 53 and cci extensions

hack-sentinel:
Has demonstration results in src/ontology/d3fend-protege.ccl.ttl and d3fend-protege.nist.ttl.  Can be kept or dropped after merge, but can poke through them with protege right away before rebuild.

NOTE: If you make build, will need to make install-deps again for pipenv ...or do update pipenv ala carte; packages added.


hack-sentinel:
Reworking to pass flake8 checks.

Issue #79, opened 2022-11-03
Artifact: Network Analysis Software


ioggstream:
## I expect

- an artifact for network analysis/scanning (Software or Tool)

netfl0:
Agree, this would be a Network Traffic Analysis Software.

PR #80, opened 2022-11-03
Fix: #79. Add NetworkTrafficAnalysisSoftware.

ioggstream:
## This PR

Add NetworkTrafficAnalysisSoftware

Fix: #79

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -6305,6 +6305,16 @@ Administrators collect information on network nodes in their architecture using 
     :definition "Analyzing intercepted or summarized computer network traffic to detect unauthorized activity." ;
     :enables :Detect .
 
+:NetworkTrafficAnalysisSoftware a :DigitalArtifact,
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -6305,6 +6305,16 @@ Administrators collect information on network nodes in their architecture using
     :definition "Analyzing intercepted or summarized computer network traffic to detect unauthorized activity." ;
     :enables :Detect .
 
+:NetworkTrafficAnalysisSoftware a :DigitalArtifact,
+        owl:Class,
+        owl:NamedIndividual ;
```
netfl0, line 6310: we are not punning in this file at this time, please remove then and then it looks great!

netfl0 (changes requested):
Only one issue.

src/ontology/d3fend-protege.ttl
ioggstream: Done!

PR #90, opened 2023-01-06
[Correction] Fixed 'version' data property and several 'has-link' instances

### LINKED ISSUE #89: [Correction] The type of "version" data property contains inconsistencies
Hi! I found inconsistency in the ontology:
1) For individuals `NIST_SP_800-53_R3`, `NIST_SP_800-53_R4`, `NIST_SP_800-53_R5`, the **version** property is represented _as an integer_. However, the class higher in the hierarchy `Control Catalog` contains the axiom that the version _is a string_. Similarly for the `Capability Implementation` class.
**Proposed solutions**:
a) Correct all values of data property "version" to the corresponding strings.
b) For the `Capability Implementation` and `Control Catalog` classes, specify the types for 'version' as a union: `(xsd:decimal or xsd:integer or xsd:string)`.
2) The domain of the data property "version" is 'Capability Implementation', however, this data property is also used in the 'Control Catalog' class. So the domain should be union: `('Capability Implementation' or 'Control Catalog')`.
3) Depending on the chosen solution in the first section:
a) Range of data property "version" set to `xsd:string`.
b) Set the range of data property "version" equal to the union: `(xsd:decimal or xsd:integer or xsd:string)`. 

If relevant, I can create a PR with one of the proposed solutions.

aw350m33d:
This Pull Request is the result of a discussion in issue https://github.com/d3fend/d3fend-ontology/issues/89.

In accordance with the approach chosen in the discussion, the following changes were made:
1) The version's `rdfs:domain` has been updated and now it is a union: `('Capability Implementation' and 'Control Catalog')`.

2) The version's `rdfs:range` has been updated and now it is `xsd:string`.

3) All numeric instances of `version` data property are converted to the corresponding strings.

4) Added a link to the patent for the individual `Reference - Using spanning tree protocol (STP) to enhance layer-2 topology maps`.

5) The types of some instances of `has-link` data propertiey have been converted to `xsd:anyURI`. 
**Nonte:** Although there is a `make-has-links-anyURI.rq` query execution stage in the pipeline of the ontology building (so the final ontology will contain the correct types), I believe that the `d3fend-protege.ttl` file itself is better maintained in a consistent state

Resolves #89

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -1463,13 +1463,11 @@ skos:altLabel a owl:AnnotationProperty ;
     rdfs:label "version"@en ;
     rdfs:subPropertyOf :d3fend-catalog-data-property,
         :d3fend-external-control-data-property ;
-    rdfs:domain :CapabilityImplementation ;
-    rdfs:range xsd:string,
-        [ a rdfs:Datatype ;
+    rdfs:domain [ a owl:Class ;
             owl:unionOf (
-                    xsd:decimal
-                    xsd:integer
-                    xsd:string ) ] ;
+                    :CapabilityImplementation
+                    :ControlCatalog ) ] ;
+    rdfs:range xsd:string ;
     rdfs:comment "x version y: The product or service x has the version y." .
 
 :windows-registry-data-property a owl:DatatypeProperty ;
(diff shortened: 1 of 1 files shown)


netfl0:
@aw350m33d please let me know if you would like to be listed as a contributor and what format.

Thank you!

aw350m33d:
> @aw350m33d please let me know if you would like to be listed as a contributor and what format.
> 
> Thank you!

@netfl0 I'll be happy to be listed as a contributor. I don't have any format requirements. The same format as other contributors will do. You can use a nickname and/or full name.

Thanks a lot!

Issue #95, opened 2023-01-19
Identifier Activity Analysis


kelby-shelton:
A new technique under Detect > Identifier Analysis which describes the process of taking known malicious Identifiers and determining the artifacts that have interacted with those Identifiers. 

Interaction with the identifier could take the form of network traffic activity, process activity, file activity, registry activity, etc.

hack-sentinel:
Initial version in, will update content (see #116) from recent work.

hack-sentinel:
#116 completed/merged.  Done.

PR #114, opened 2023-02-22
Review and pull for feature #95; add identifier activity analysis

hack-sentinel:
Please review as part of pull request; currently a clean merge. 

1) Review kb-articles.  They are small but may not have properly. captured verbal discussion between PK and KS.

2) Without adding adding something like an Identifier Activity class, which would really be a broad set of log/data record types, or "analyzes log" to get after that idea another way, simplest solution was to have this just be a another case of analyzes->Identifier* as the core relationship. (otherwise this would be non-parallel with sibling and parent class here).  *True for all subclasses here*, so added restriction at top Identifier Analysis level.


hack-sentinel:
Need to fold in reference and kb-article text/content from PPT slides that were done in parallel (netfl0 et al).  See #116.

PR #152, opened 2023-04-10
Specific Syscalls 

ryantxu1:
Adding specific linux and windows syscalls to the taxonomy.

Before merging @hack-sentinel I have a few questions:
- I don’t know how to get hierarchy tree updated
- Do these entries of specific calls need neighbors?
- I have a case where the Linux Pause Syscall can suspend a thread and a process. How can I duplicate a "Linux Pause" class in protégé? If I can't do that, how should the protégé iri be handled?

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -3929,6 +3929,14 @@ A credential may be:
             owl:onProperty :claims ;
             owl:someValuesFrom :DefensiveTechnique ] .
 
+:DeleteFile a owl:Class ;
+    rdfs:label "Delete File" ;
+    rdfs:subClassOf :SystemCall,
+        [ a owl:Restriction ;
+            owl:onProperty :deletes ;
+            owl:someValuesFrom :File ] ;
+    :definition "Remove a file from a machine." .
+
 :Dependency a owl:Class ;
     rdfs:label "Dependency" ;
     rdfs:subClassOf :DigitalArtifact,
@@ -5461,6 +5469,24 @@ A homoglyph, in this context, is a deceptive string or word which looks like a t
     :definition "A document file encoded in HTML.The HyperText Markup Language, or HTML is the standard markup language for documents designed to be displayed in a web browser. It can be assisted by technologies such as Cascading Style Sheets (CSS) and scripting languages such as JavaScript. Web browsers receive HTML documents from a web server or from local storage and render the documents into multimedia web pages. HTML describes the structure of a web page semantically and originally included cues for the appearance of the document." ;
     rdfs:seeAlso <http://dbpedia.org/resource/HTML> .
 
+<http://d3fend.mitre.org/ontologies/d3fend.owl#LinuxOpenArgumentO_RDONLY,O_WRONLY,O_RDWR> a owl:Class ;
+    rdfs:label "Linux Open Argument O_RDONLY, O_WRONLY, O_RDWR" ;
+    rdfs:subClassOf :OpenFile ;
+    :definition "Opens a file specified by pathname." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/open.2.html" .
+
+<http://d3fend.mitre.org/ontologies/d3fend.owl#LinuxOpenAt2ArgumentO_RDONLY,O_WRONLY,O_RDWR> a owl:Class ;
+    rdfs:label "Linux OpenAt2 Argument O_RDONLY, O_WRONLY, O_RDWR" ;
+    rdfs:subClassOf :OpenFile ;
+    :definition "Extension of Linux Openat." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/openat2.2.html" .
+
+<http://d3fend.mitre.org/ontologies/d3fend.owl#LinuxOpenAtArgumentO_RDONLY,O_WRONLY,O_RDWR> a owl:Class ;
+    rdfs:label "Linux OpenAt Argument O_RDONLY, O_WRONLY, O_RDWR" ;
+    rdfs:subClassOf :OpenFile ;
+    :definition "Same functionality as Linux Open but slight differences in parameter." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/openat.2.html" .
+
 :HumanInputDeviceFirmware a owl:Class ;
     rdfs:label "Human Input Device Firmware" ;
     rdfs:subClassOf :PeripheralFirmware ;
@@ -6270,6 +6296,12 @@ Most current Unix-like systems and Microsoft Windows support loadable kernel mod
     rdfs:subClassOf :DigitalArtifact ;
     rdfs:seeAlso "https://dbpedia.org/resource/Link" .
 
+:Linux_Exit a owl:Class ;
+    rdfs:label "Linux _Exit" ;
+    rdfs:subClassOf :TerminateProcess ;
+    :definition "Terminate the calling process." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/exit.2.html" .
+
 :LinuxClone a owl:Class ;
     rdfs:label "Linux Clone" ;
     rdfs:subClassOf :CreateProcess ;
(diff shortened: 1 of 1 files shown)


ioggstream:
Shouldn't we have a POSIX taxonomy first?

netfl0:
> Shouldn't we have a POSIX taxonomy first?

This is a great idea, since we've already done the windows analysis perhaps second :) 

@ryantxu1 let me know your thoughts here.

ryantxu1:
@netfl0 Open to exploring this eventually, are we envisioning a similar process of bucketing the system interfaces in POSIX similar to the linux/windows syscalls?

src/ontology/d3fend-protege.ttl
```diff
@@ -15914,6 +16166,131 @@ User web session data is collected over a period of time to create a user behavi
...
+:WindowsNtReadFile a owl:Class ;
+    rdfs:label "Windows NtReadFile" ;
+    rdfs:subClassOf :ReadFile ;
+    rdfs:seeAlso "http://undocumented.ntinternals.net/UserMode/Undocumented%20Functions/NT%20Objects/File/NtReadFile.html" .
+
+:WindowsNtReadFileScatter a owl:Class ;
+    rdfs:label "Windows NtReadFileScatter" ;
+    rdfs:subClassOf :ReadFile ;
+    :definition "Reads specified block from file into multiple buffers. Each buffer must have one page length." ;
+    rdfs:seeAlso "http://undocumented.ntinternals.net/UserMode/Undocumented%20Functions/NT%20Objects/File/NtReadFileScatter.html" .
```
netfl0, line 16261: should be rdfs:isDefinedBy

src/ontology/d3fend-protege.ttl
```diff
@@ -16216,8 +16593,7 @@ order to constitute a complete standard. For a complete definition of all requir
     rdfs:label "Linux ELF File 64bit" .
 
 :LinuxExec a :CreateProcess,
```
ioggstream, line 16595: LinuxExec does not create a process. Instead, it replaces the  program that is currently being run by the calling process to be replaced with a new program, with newly initialized stack, heap, and (initialized and
       uninitialized) data segments.

There are other parts of the process that are not replaced, eg.,

The process's real UID and real GID, as well its supplementary group IDs, are unchanged;
file descriptors remain open unless marked close-on-exec.

Syscalls are a slippery slope...

src/ontology/d3fend-protege.ttl
```diff
@@ -6264,6 +6296,216 @@ Most current Unix-like systems and Microsoft Windows support loadable kernel mod
     rdfs:subClassOf :DigitalArtifact ;
     rdfs:seeAlso "https://dbpedia.org/resource/Link" .
 
+:Linux_Exit a owl:Class ;
+    rdfs:label "Linux _Exit" ;
```
ioggstream, line 6300: Can't find _exit in `man 2 syscalls | grep _exit`

Here and elsewhere: why don't use the verbatim syscall name, e.g. "exit(2)"  in the label?

src/ontology/d3fend-protege.ttl
```diff
@@ -6264,6 +6296,216 @@ Most current Unix-like systems and Microsoft Windows support loadable kernel mod
...
+    rdfs:label "Linux Execve" ;
+    rdfs:subClassOf :CreateProcess ;
+    :definition "Execute program." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/execve.2.html" .
+
+:LinuxExecveat a owl:Class ;
+    rdfs:label "Linux Execveat" ;
+    rdfs:subClassOf :CreateProcess ;
+    :definition "Execute program relative to a directory file descriptor." ;
+    rdfs:seeAlso "https://man7.org/linux/man-pages/man2/execveat.2.html" .
```
ioggstream, line 6353: Are Execute and Create in the same class?

netfl0: They both Create Process (this they are both a sublcass of that semantic class)  if that makes sense. We've got some clean up coming, @ryantxu1 will be pushing to this branch soon.

ioggstream: Understood.  Since I don't think this reflects the Linux taxonomy I will wait for the cleanup.

src/ontology/d3fend-protege.ttl
```diff
@@ -5455,6 +5469,24 @@ A homoglyph, in this context, is a deceptive string or word which looks like a t
     :definition "A document file encoded in HTML.The HyperText Markup Language, or HTML is the standard markup language for documents designed to be displayed in a web browser. It can be assisted by technologies such as Cascading Style Sheets (CSS) and scripting languages such as JavaScript. Web browsers receive HTML documents from a web server or from local storage and render the documents into multimedia web pages. HTML describes the structure of a web page semantically and originally included cues for the appearance of the document." ;
     rdfs:seeAlso <http://dbpedia.org/resource/HTML> .
 
+<http://d3fend.mitre.org/ontologies/d3fend.owl#LinuxOpenArgumentO_RDONLY,O_WRONLY,O_RDWR> a owl:Class ;
```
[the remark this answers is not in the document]
netfl0: The detection analytic developers are the ones who study those to look for useful patterns. Unless you are actually developing an analytic, you would care about the D3FEND technique that might exist in a product that is making sense of these things.

netfl0:
let me know when ready for review

src/ontology/d3fend-protege.ttl
ryantxu1: @ioggstream  I put '_exit' because the man page of that specifically states that as the name when actually calling the function.
https://man7.org/linux/man-pages/man2/exit.2.html

That is also why I left out the '(2)' in the labels as I was focusing on the function names themselves. However, I'm happy to reconsider the naming convention if there's a distinct advantage to do so!

ryantxu1:
@netfl0 Ready for review

ryantxu1:
@netfl0 Ready for review

src/ontology/d3fend-protege.ttl
```diff
@@ -4630,6 +4640,17 @@ On a Windows machine the Windows Defender Application Control (WDAC) policy enfo
     :definition "An executable script is written in a scripting language and interpreted at run time. This is in contrast with an executable binary, which contains machine code instructions for a physical CPU or byte code for a virtual machine." ;
     rdfs:seeAlso <http://dbpedia.org/resource/Executable> .
 
+:ExecuteProcess a owl:Class ;
+    rdfs:label "Execute Process" ;
+    rdfs:subClassOf :SpawnProcess,
+        [ a owl:Restriction ;
+            owl:onProperty :executes ;
+            owl:someValuesFrom :Process ] ;
+    :definition "Executes a process." ;
```
ioggstream, line 4649: IIUC, you execute a program, not a process :)

src/ontology/d3fend-protege.ttl
```diff
@@ -4525,7 +4535,7 @@ SafeSEH might be applied only to some executable files or modules, allowing an a
             owl:someValuesFrom :ExecutableFile ],
         [ a owl:Restriction ;
             owl:onProperty :restricts ;
-            owl:someValuesFrom :CreateProcess ] ;
+            owl:someValuesFrom :SpawnProcess ] ;
```
ioggstream, line 4538: Formally, when the execve(2) is invoked, the process was already spawn.

Moreover, I can create a new process just forking an existing one without execve(2) - e.g., running another instance of the same program.

src/ontology/d3fend-protege.ttl
ioggstream: Using system call names or function signatures is a "design" choice. Since implementations can vary in time (see also the link you posted https://man7.org/linux/man-pages/man2/exit.2.html) if we want to focus on Linux, I'd use a man-like uri (e.g. `man-pages/man2/exit.2.html` references both `exit` and `2`).

This allows us to define a POSIX version (e.g., `POSIX_exit`)

src/ontology/d3fend-protege.ttl
```diff
@@ -9526,6 +9775,15 @@ Application-layer discovery:
     rdfs:subClassOf :TechniqueReference ;
     :pref-label "Source Code" .
 
+:SpawnProcess a owl:Class ;
+    rdfs:label "Spawn Process" ;
+    skos:altLabel "Process Spawn" ;
+    rdfs:subClassOf :SystemCall ;
+    :definition "A process spawn refers to a function that loads and executes a new child process.The current process may wait for the child to terminate or may continue to execute asynchronously. Creating a new subprocess requires enough memory in which both the child process and the current program can execute. There is a family of spawn functions in DOS, inherited by Microsoft Windows. There is also a different family of spawn functions in an optional extension of the POSIX standards.  Fork-exec is another technique combining two Unix system calls, which can effect a process spawn." ;
```
ioggstream, line 9782:
```suggestion
    :definition "A process spawn refers to a function that loads an executable and executes it in a new child process. The current process may wait for the child to terminate or may continue to execute asynchronously. Creating a subprocess requires enough memory in which both the child process and the current program can execute. There is a family of spawn functions in DOS, inherited by Microsoft Windows. There is also a different family of spawn functions in an optional extension of the POSIX standards.  Fork-exec is another technique combining two Unix system calls, which can effect a process spawn." ;
```

src/ontology/d3fend-protege.ttl
```diff
@@ -9728,9 +9986,18 @@ Extremely complex password requirements may lead users to saving passwords in te
...
             owl:someValuesFrom :Process ] .
 
+:SuspendThread a owl:Class ;
+    rdfs:label "Suspend Thread" ;
+    rdfs:subClassOf :SystemCall,
+        [ a owl:Restriction ;
+            owl:onProperty :suspends ;
+            owl:someValuesFrom :Thread ] ;
+    :definition "Suspending a thread causes the thread to stop executing user-mode code." ;
+    rdfs:seeAlso "https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-suspendthread" .
```
ioggstream, line 9999: See also https://man7.org/linux/man-pages/man2/signal.2.html

Issue #184, opened 2023-06-13
Add CWE taxonomy from children of pillars down to support D3FEND / CWE mapping queries


hack-sentinel:
- [x] Add CWE classes hierarchy from pillars on down and create puns; unify under Weakness; create subClassOf for ChildOf relations.

netfl0:
this done?

netfl0:
I'll take that as a yes.

PR #191, opened 2023-06-22
#184. To complete CWE work so query would run as intended, added missing rdfs:labels on weakness properties.

hack-sentinel:
Needed rdfs:labels for weakness properties


netfl0 (approved):
looks good

Issue #195, opened 2023-08-30
Browser-based Process Isolation 


mdoherty-axiom:
Many modern applications use web browsers as runtime environments. For security, Chrome and Firefox have implemented extensive process isolation techniques, including cross-origin blocking, site isolation, cookie and storage isolation, and sandboxing. 

While Kernel-based Process Isolation is used to enforce some of these protections, most are implemented within the browser itself. Browser-based Process Isolation has a unique set of vulnerabilities and attacks that cannot be modeled with Kernel-based Process Isolation or Hardware-based Process Isolation.

PR #196, opened 2023-08-30
Browser isolation

mdoherty-axiom:
This Pull Request attempts to resolve Issue #195 by adding Browser-based Process Isolation as a D3FEND technique under Execution Isolation. We have added references to the relevant Chrome and Firefox security architectures.

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -26124,7 +26124,7 @@ This patent describes ensuring that a driver associated with the agent is initia
     :kb-reference-title "Integrity assurance through early loading in the boot phase" .
 
 :Reference-IntroducingFirefoxNewSiteIsolationArchitecture a :InternetArticleReference,
-        owl:NamedIndividual,
+        owl:NamedIndividual ;
     rdfs:label "Reference - Introducing Firefox's new Site Isolation Architecture" ;
     :has-link "https://hacks.mozilla.org/2021/05/introducing-firefox-new-site-isolation-security-architecture/" ;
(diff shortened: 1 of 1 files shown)


netfl0:
Thank you for this pull request! We've been playing a lot of catch up so we appreciate your patience.

After reviewing the references, it appears we are talking about two distinct technologies. 

The first [example with firefox](https://hacks.mozilla.org/2021/05/introducing-firefox-new-site-isolation-security-architecture/) uses multiple processes, one privileged, and an API by which child/risky processes make requests to the privileged one for things like file write/access. Here is a key quote from the writeup:

> With Site Isolation, Firefox loads each site in its own process, thereby isolating their memory from each other, and relies on security guarantees of the operating system.

They don't specify what security guarantees, presumably they are using some sort of IPC between the processed but we'd need to do more research, it may use shared memory, not sure.

The second is a remote (network sense) browsing session ala some sort of remote frame buffer type of mechanism.


This should be broken into two separate techniques, (need a better names).
1. IPC-based Process Isolation (This is more of a design pattern to some degree. It is in opposition to just using threads in the same process as in Firefox, i am not sure what mechanisms they are using to enforce permissions)
2. Remote Application, windows called this [RemoteApp](https://techcommunity.microsoft.com/t5/security-compliance-and-identity/introducing-remoteapp-and-desktop-connections/ba-p/246803) at one time.

I'll recommend some new names for those two, then you can update the pull request accordingly.

netfl0:
Work ported to #320. Excellent contribution thank you @mdoherty-axiom @vikhari-axiom

Issue #215, opened 2024-02-08
Add Process Injection Artifacts


ryantxu1:
Adding digital artifacts related to process injection to link to diagrams for our upcoming blog post.

PR #217, opened 2024-02-13
Process Injection Syscalls

ryantxu1:
Addresses issue #215


ryantxu1:
@netfl0 @hack-sentinel if yall, have time can you please review? One thing I was unsure about was 

Get Thread Context: analyzes some Thread??

Issue #202, opened 2023-09-22
System Shutdown / Reboot


apapaa:
**Note: all sections are required.**

Evict > Hardware Eviction > System Shutdown / Reboot
> The name should indicate which digital artifacts are in play, and what actions are applied to those artifacts.

## Digital Artifacts
> What are the relevant [D3FEND Digital Artifacts](https://d3fend.mitre.org/dao/) to this new technique, please propose new artifacts if you cannot find them in D3FEND.

[d3f:HardwareDevice](https://d3fend.mitre.org/dao/artifact/d3f:HardwareDevice)
[d3f:Process](https://d3fend.mitre.org/dao/artifact/d3f:Process)

## Definition
> One or two sentence definition in the style of other d3fend techniques.

In a similar way to how an attacker can use System Shutdown / Reboot [T1529](https://attack.mitre.org/techniques/T1529/) to disrupt a victim, defenders may use this technique against attackers who have compromised hardware systems to either evict them or protect the particular system from suffering any further damage.

## How it works
> Section explaining how the technique works.

System Shutdown / Reboot can either be initiated in the physical presence of the device using the power functions or remotely using the provided user interface or an installed EDR agent (with the available function). Compromised systems may however not respond to remote commands to shutdown or reboot, requiring physical intervention.

## Considerations
> What should people know about this technique, pros/cons, pitfalls etc.

- Shutting down a system will usually result in the memory losing its state which can be useful in forensic activities so this should be considered when deciding to shutdown.
- If the attacker has maintained persistence by altering the bootup disk, flashing the ROM or injecting their process into the startup of the system, the reboot technique may not be effective in evicting the attacker however it may buy you time in isolating the system from the rest of the network.
- Shutting down or rebooting systems may disrupt access to computer resources for legitimate users.

## References
> High quality publicly available technical documents.

https://attack.mitre.org/techniques/T1529/

netfl0:
Thank you for the contribution, we agree we should add this. 

A reboot mostly affects running processes ( and their related memory artifacts). We should model this against those.

apapaa:
> Thank you for the contribution, we agree we should add this.
> 
> A reboot mostly affects running processes ( and their related memory artifacts). We should model this against those.

I've added the "Process" artefact but I think our definition needs to be expanded to be similar to ATT&CK definition to include "memory" space...

D3FEND Process Definition:
[...]

PR #219, opened 2024-02-20
Adding System Shutdown/Reboot Technique

ryantxu1:
Addressing #202


netfl0 (approved):
Great work, thank you.

PR #224, opened 2024-03-25
DRAFT: Add MS matrices extensions

aw350m33d:
Changes have been made as instructed and the changes are reflected below.

## **Added new ATTACK-like matrices**
New extensions `atrm` and `tmfk` were created. These extensions contain Python code to generate corresponding ontologies from STIX bundles.

## **Renamed existing concept to reflect its general usage**
The `d3f:OffensiveTechnique` class has been renamed to `d3f:ATTACKOffensiveTechnique`, following the example of the `d3f:ATTACKMitigation` class. Additionally, a 'new' generic concept `d3f:OffensiveTechnique` was introduced. That class can represent any offensive techniques across different frameworks or methodologies.

## **Created new generic container classes for ATTACK-like concepts**
A new high level classes has been introduced in the ontology: `d3f:TTPClassifier` and `d3f:Matrices`. This classes can be used to organize and group ATTACK-like concepts.

## **Updated report for missing techniques' ids**
The query `missing-off-tech-artifacts.rq` has been updated to include the new OPTIONAL parts with `atrm-id` and `tmfk-id`.

A minor typo in the `missing-off-tech-artifacts.rq` file has been corrected, where the `pentity` was incorrectly referred to as `p_entity `.


netfl0:
@aw350m33d thank you for this contribution. We have decided that this should be maintained as a separate project for a variety of reasons, namely that we are not sure how the degree to which these non-mitre matrices will be maintained over time and the level of investment therein.

I noticed the CI/CD job is failing on this, is that working for you?

I'd recommend setting up this as a fork that inherits from D3FEND Upstream with some CI/CD on it to automatically keep it up to date. We can gauge interest over time and revisit our decision in the future based on community engagement and interest.

We'd be happy to discuss further and are thrilled about this body of work you put together.

aw350m33d:
@netfl0 thank you for your feedback and for the time you spent analyzing my proposal!

I didn’t dive into troubleshooting pipeline errors until I received your approval of the proposed ideas.

I'll try to implement these ideas as a separate project that references D3FEND through submodules. Maintaining it as a fork most likely won’t work, since I made changes (namely, renaming d3f:OffensiveTechnique to d3f:ATTACKOffensiveTechnique) that would be difficult to automatically synchronize.

PR #228, opened 2024-04-16
Ryantxu1/proc inj modeling

ryantxu1:


src/ontology/d3fend-protege.ttl
```diff
@@ -14749,213 +14747,216 @@ Operating System Monitoring Techniques have varied implementations including bui
...
-    rdfs:subClassOf :OSAPIFunction,
+    rdfs:subClassOf :OSAPISystemFunction,
         [ a owl:Restriction ;
             owl:onProperty :invokes ;
             owl:someValuesFrom :FreeMemory ] .
 
 :OSAPIFunction a owl:Class ;
     rdfs:label "OS API Function" ;
-    rdfs:subClassOf :Software ;
+    rdfs:subClassOf :Subroutine ;
```
netfl0: I thought we agreed to move this keep this as a peer to subroutine.

The subroutine taxonomy is meant to be "Semantic Subroutine". These OS API System functions could be also sublclassed under their specific subroutine. E.g. **OS API Free Memory** would be a type of **Memory Free Function**. This is in contrast to a separate tree under subroutine. 

I am open to changing the name of Subroutine to Semantic Subroutine, but for now, i am adding a comment to d3f:Subroutine to explain this was my design intent with the Subroutine taxonomy. (the use case for this taxo is around source code quality analyzers). I'll also add a synonym **Semantic Subroutine**

If we need a unifying subroutine class for some reason, we will add it then.

CC @hack-sentinel

netfl0 (commented):
I will push some updates to the pull request.

Issue #227, opened 2024-04-15
Package URL 


aamedina:
# d3f:PackageURL
I'm not exactly sure how to best model this yet, but I think it's important to have a way to represent package URLs and be able to make queries based on the type of the package (Maven, NPM, etc.) and its data components (namespace, name, version, qualifiers, subpath).

Before I open a PR for this I wanted to get feedback on the approach I'm taking and if this would be useful to others. I plan on using these properties to annotate software composition analysis with D3FEND and would like to use PURLs to uniquely identify software packages across databases.

## Definition

```ttl
@prefix d3f: <http://d3fend.mitre.org/ontologies/d3fend.owl#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

d3f:PackageURL a owl:Class ;
    rdfs:label "Package URL" ;
    rdfs:subClassOf d3f:Identifier, 
        [ a owl:Restriction ;
          owl:onProperty d3f:identifies ;
          owl:someValuesFrom d3f:SoftwarePackage
        ],
        [ a owl:Restriction ;
          owl:onProperty d3f:package-type ;
          owl:someValuesFrom d3f:PackageURLType
        ] ;
        [ a owl:Restriction ;
          owl:onProperty d3f:package-namespace ;
          owl:someValuesFrom xsd:string
        ],
        [ a owl:Restriction ;
          owl:onProperty d3f:package-name ;
          owl:someValuesFrom xsd:string
        ],
        [ a owl:Restriction ;
          owl:onProperty d3f:package-version ;
          owl:someValuesFrom xsd:string
        ],
        [ a owl:Restriction ;
          owl:onProperty d3f:package-qualifiers ;
          owl:someValuesFrom xsd:string
        ],
        [ a owl:Restriction ;
          owl:onProperty d3f:package-subpath ;
          owl:someValuesFrom xsd:string
        ] ;
    d3f:definition """purl stands for package URL.

A purl is a URL composed of seven components:

scheme:type/namespace/name@version?qualifiers#subpath

Components are separated by a specific character for unambiguous parsing.

The definition for each components is:

* scheme: this is the URL scheme with the constant value of \"pkg\". This is not modeled in the RDF representation as it is tautological.
[...]

PR #243, opened 2024-05-02
Using package URLs to identify some SoftwarePackage

aamedina:
Addresses #227 

* [rdfs:seeAlso](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst).
* The d3f:definition is sourced from OCSF "package" object dictionary property "purl" via [rdfs:isDefinedBy](https://schema.ocsf.io/objects/package).

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -1014,6 +1014,18 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:label "publishes" ;
     rdfs:subPropertyOf :d3fend-catalog-object-property .
 
+:purl a owl:InverseFunctionalProperty,
+        owl:ObjectProperty ;
+    rdfs:label "has package URL" ;
+    skos:altLabel "purl" ;
+    rdfs:subPropertyOf :identified-by ;
+    rdfs:domain :SoftwarePackage ;
+    rdfs:range :URL ;
+    rdfs:comment "A purl is a URL string used to identify and locate a software package in a mostly universal and uniform way across programming languages, package managers, packaging conventions, tools, APIs and databases." ;
+    rdfs:isDefinedBy <https://schema.ocsf.io/objects/package> ;
+    :definition "x has package URL y: The subject software package x is identified by the package URL y." ;
+    rdfs:seeAlso <https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst> .
+
 :queries a owl:ObjectProperty ;
     rdfs:label "queries" ;
     rdfs:subPropertyOf :associated-with,


src/ontology/d3fend-protege.ttl
```diff
@@ -15153,6 +15153,33 @@ The OWL languages are characterized by formal semantics. They are built upon the
 1. Web Ontology Language. (2023, April 23). In _Wikipedia_. [Link](https://en.wikipedia.org/wiki/Web_Ontology_Language)""" ;
     :synonym "Web Ontology Language" .
 
+:PackageURL a owl:Class ;
+    rdfs:label "Package URL" ;
+    rdfs:subClassOf :Identifier,
```
netfl0, line 15158: should be subclass of d3f:URL

src/ontology/d3fend-protege.ttl
```diff
@@ -15153,6 +15153,33 @@ The OWL languages are characterized by formal semantics. They are built upon the
...
+        [ a owl:Restriction ;
+            owl:onProperty :identifies ;
```
netfl0, line 15160: should be :addresses

(for consistency. we may improve this in future)

src/ontology/d3fend-protege.ttl
```diff
@@ -15153,6 +15153,33 @@ The OWL languages are characterized by formal semantics. They are built upon the
...
+            owl:someValuesFrom :SoftwarePackage ] ;
+    rdfs:isDefinedBy <https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst> ;
```
netfl0, line 15162: should be rdfs:seeAlso , not defined by. We must to keep it semantic.

src/ontology/d3fend-protege.ttl
aamedina: Should we update the definition of d3f:URL then? This is what informed my modeling decision:

```turtle
:URL a owl:Class ;
    rdfs:label "URL" ;
    skos:altLabel "Uniform Resource Locator" ;
    rdfs:subClassOf :Identifier,
        [ a owl:Restriction ;
            owl:onProperty :addresses ;
            owl:someValuesFrom :Resource ] ;
    rdfs:isDefinedBy <http://dbpedia.org/resource/Uniform_Resource_Locator> ;
    :definition "A Uniform Resource Locator (URL), commonly informally termed a web address (a term which is not defined identically) is a reference to a web resource that specifies its location on a computer network and a mechanism for retrieving it.A URL is a specific type of Uniform Resource Identifier (URI), although many people use the two terms interchangeably. A URL implies the means to access an indicated resource, which is not true of every URI. URLs occur most commonly to reference web pages (http), but are also used for file transfer (ftp), email (mailto), database access (JDBC), and many other applications." ;
    rdfs:seeAlso "https://schema.ocsf.io/objects/url" .
 ```
 
 The definition (including the semantics) implies that the URL is meant to access the indicated resource, which isn't the case with purls. It is meant as the identity of some package.

File: src/ontology/d3fend-protege.ttl [hunk 4]
```diff
@@ -15153,6 +15153,33 @@ The OWL languages are characterized by formal semantics. They are built upon the
...
+    :definition """purl stands for package URL.
```
netfl0, line 15163: This definition needs to be abstracted. It should be concise. This would be defining a concept d3f:PURLPackageURL if that makes sense. This would be a very specific type of package url.

src/ontology/d3fend-protege.ttl
aamedina: OK. My thinking was in this case I think the PURL spec truly is "defining" the concept. Will update.

src/ontology/d3fend-protege.ttl
aamedina: I know what you mean, in that case I would prefer it be called d3f:PURL I think.

src/ontology/d3fend-protege.ttl
aamedina: I am not sure that a purl addresses a package. 

```
x addresses y: Relates a pointer x to a digital artifact y located in the address space to which x points. The address space is part of some digital store, whether it be in memory, an image, or a persistent storage device.
```

It does identify a package. It may be associated with some artifact which is loaded into memory and would then become addressable (like the file:// pointing to a specific DLL in a binary analysis, which was compiled by the package identified by a PURL, and then now its code and data is loaded into memory and becomes an addressable resource, but that is a different artifact than the PURL that identifies its source).

netfl0: I was hinting we need to improve this :) This is how we use URL at the moment. 

However, you are inspiring me to fix this now however.

https://d3fend.mitre.org/dao/artifact/d3f:URL/

There is a distinction between Locator and Identifier. 

:locates
:references
:identifies

We need something to indicate the **resolvability** aspect.

aamedina: I admit I am kind of confused by the definition of PURL as a URL myself. I personally use it to identify a software package uniquely as a IRI, but the OCSF definition and the spec documented I linked do indicate that it can be used to "locate" the package as well as identify it. But semantically, PURLs can't control how package managers work, which would actually perform the resolution, I mean it would be nice if all package managers converged on a universal locator, but they are different concepts. It reminds me more of how a file hash is used to identify files. You can have many file hashes that identify the same file. A purl in my interpretation should unambiguously refer to some software package, but the actual identifier is just a string. You can have many PURLs refer to the same software package, with different qualifiers for example. Will continue to think about this.

aamedina:
I removed the :PURL class and instead reverted to a previous idea which was to use a inverse functional property to identify some software package by URL. Then we don't need a new class for the "Package URL". 

Let me know your thoughts on this approach. 

```turtle
:purl a owl:InverseFunctionalProperty,
        owl:ObjectProperty ;
    rdfs:label "has package URL" ;
    skos:altLabel "purl" ;
    rdfs:subPropertyOf :identified-by ;
    rdfs:domain :SoftwarePackage ;
    rdfs:range :URL ;
    rdfs:comment "A purl is a URL string used to identify and locate a software package in a mostly universal and uniform way across programming languages, package managers, packaging conventions, tools, APIs and databases." ;
    rdfs:isDefinedBy <https://schema.ocsf.io/objects/package> ;
    :definition "x has package URL y: The subject software package x is identified by the package URL y." ;
    rdfs:seeAlso <https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst> .
 ```

netfl0:
We need to systematically map in OCSF properties. This has been a plan in the works for some time, but we'll start in earnest in 2025. This particular property, purl, is doing to be important to the SBOM initiatives so we are going to want to synchronize with them for compatibility. I am going to close this pull request for now and we will pick it back up in 2025 with ideally some sort of semi-automated approach to sync D3FEND data and object properties to OCSF properties and field names, it will require some sort of mapping file/code.

Issue #230, opened 2024-04-16
d3f:unloads


aamedina:
# d3f:unloads
I would like to be able to reason over facts asserting that some individual artifact (like a host d3f:Kernel executing a process being traced) unloads a d3f:KernelModule or d3f:HardwareDriver. 

```ttl
d3f:unloads a owl:ObjectProperty ;
  rdfs:label "unloads" ;
  rdfs:subPropertyOf :evicts ;
  d3f:definition "x unloads y: The technique or artifact performs the action of unloading some artifact (applications, kernel modules, or hardware drivers, etc.) from a computer's memory." .
```

Do these system calls make sense? They don't fit nicely into the OSAPIFunction hierarchy, but they are system calls in Linux.

```ttl
d3f:LoadModule a owl:Class ;
  d3f:definition "A system call that loads a driver or extension into the kernel." ;
  rdfs:seeAlso <https://man7.org/linux/man-pages/man2/init_module.2.html> ;
  rdfs:subClassOf d3f:SystemCall,
    [ a owl:Restriction ;
      owl:onProperty d3f:loads ;
      owl:someValuesFrom [
        owl:unionOf ( d3f:KernelModule d3f:HardwareDriver )
        ] ;
    ] .

d3f:UnloadModule a owl:Class ;
  d3f:definition "A system call that unloads a driver or extension from the kernel." ;
  rdfs:seeAlso <https://man7.org/linux/man-pages/man2/delete_module.2.html> ;
  rdfs:subClassOf d3f:SystemCall,
    [ a owl:Restriction ;
      owl:onProperty d3f:unloads ;
      owl:someValuesFrom [
        owl:unionOf ( d3f:KernelModule d3f:HardwareDriver )
        ] ;
    ] .
```

# References
* https://schema.ocsf.io/classes/kernel_extension
* https://github.com/mitre-attack/car/blob/master/data_model/module.yaml
* https://github.com/mitre-attack/car/blob/master/data_model/driver.yaml

netfl0:
Those make sense to me. The OS API taxo is parallel to the system call library, we could add those as well. We will be publishing a blog on syscalls soon to explain our design. CC @ryantxu1 

The key distinction to consider is what is being hooked where.

ryantxu1:
[...]

PR #244, opened 2024-05-03
Add d3f:unloads property and associated module system calls 

aamedina:
Addresses #230 

This pull request introduces a new `d3f:unloads` property and adds related system call definitions to the ontology for handling the loading and unloading of modules, specifically targeting operations in a computer's memory.

Key changes include:
1. **New Object Property**: The `d3f:unloads` is added as an object property, defined as a technique or artifact that performs the action of unloading artifacts like kernel modules, or hardware drivers from a computer's memory. 
   
2. **New Classes**:
   - `LinuxDeleteModule`: Represents a system call to remove an unused loadable kernel module. It includes an exit function executed prior to the module's unloading.
   - `LinuxInitModule`: Describes a system call that loads an ELF image into kernel space and initializes module parameters.
   - `LoadModule` and `UnloadModule`: Define system calls for loading and unloading drivers or extensions into/from the kernel.

3. **API Definitions**:
   - `OSAPILoadModule` and `OSAPIUnloadModule`: Classes are defined to represent the operating system API functions that facilitate module loading and unloading.

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -1142,6 +1142,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:seeAlso <http://wordnet-rdf.princeton.edu/id/00354493-v> ;
     :synonym "aborts" .
 
+:unloads a owl:ObjectProperty ;
+    rdfs:label "unloads" ;
+    rdfs:subPropertyOf :evicts ;
+    :definition "x unloads y: The technique or artifact performs the action of unloading some artifact (applications, kernel modules, or hardware drivers, etc.) from a computer's memory." .
+
 :unmounts a owl:ObjectProperty ;
     rdfs:label "unmounts" ;
     rdfs:subPropertyOf :associated-with ;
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -25698,6 +25739,16 @@ Intro to Active Learning. inovex Blog.  [Link](https://www.inovex.de/de/blog/int
     rdfs:subClassOf :FileSystemLink ;
     :definition "A Unix link is a file link in a Unix file system." .
 
+:UnloadModule a owl:Class ;
+    rdfs:subClassOf :SystemCall,
+        [ a owl:Restriction ;
+            owl:onProperty :unloads ;
+            owl:someValuesFrom [ a owl:Class ;
```
netfl0, line 25746: To follow the style we have been using, can you convert to two separate someValuesFrom restrictions?

aamedina: Sure--if it's semantically equivalent in this case. The intention is to create a logical OR. PR updated.

PR #274, opened 2024-07-26
192: active digital system discovery

deedayal:


netfl0:
This is currently covered by: https://d3fend.mitre.org/technique/d3f:NetworkNodeInventory/

PR #294, opened 2024-09-16
Add relation that T1021 depends on T1078 (#292)

### LINKED ISSUE #292: d3f:T1021 should have a semantic relation to d3f:T1078
My perspective is that since all of the T1021 techniques use valid accounts (t1078) to actually perform lateral movement, there should be some kind of semantic relation from T1021 and T1078 that can be inferred through its subtechniques.

Effectively I propose a semantic relation that T1021 is related to T1078.

aamedina:
fixes #292

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -18983,6 +18983,9 @@ When system firmware verification fails a set of predefined responses is typical
     rdfs:label "Remote Services" ;
     rdfs:subClassOf :LateralMovementTechnique,
         [ a owl:Restriction ;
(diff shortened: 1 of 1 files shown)


aamedina:
Since there are no other uses of :dependsOn in the ontology (before this...), what about changing its name to d3f:depends-on for consistency in the naming conventions for properties? Needs a New PR?

netfl0:
Thank you, we do need this. 

However, we are going to need to reify dependencies.  The initial start as of last release can be seen as d3f:Condition. 

We have to consider the reqs from stakeholders on the engineering which we're still working on.

PR #305, opened 2024-10-04
cleanup anonymous individuals, fix network model.

netfl0:


ryantxu1:
Changes look ok, I think we can revisit user account deletion in Linda's stuff?

PR #319, opened 2024-11-25
Add Cloud Service Provider subclass

tareq-alkhatib:
Honestly, this is mostly an excuse to try writing something in TTL.

This change adds the `CloudServiceProvider` as a subclass of `ServiceProvider`. This is useful for anyone who wants to create CSP specific resources like AWS EC2 or Google Cloud Compute Engine. 

I understand not wanting to have AWS, Azure, or GCP as resources in an otherwise agnostic ontology, but we should have a resource to subclass for anyone who wants to extend the ontology for other uses.

I note that the ontology also does not have a `CloudService` resource either. I thought about adding it and its own subclasses (IAAS, PAAS, and SAAS) but then I would need to relate it to the aforementioned `CloudServiceProvider` as well as all of T1021.007 (Cloud Services), not to mention a whole lot of other cloud resources, at which point it felt like too complex a change for a first contribution.

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -3430,7 +3430,7 @@ Profiling request and response payloads across multiple clients to a single serv
     rdfs:label "Cloud Configuration" ;
     skos:altLabel "Cloud Configuration Information" ;
     rdfs:subClassOf :ConfigurationResource ;
-    :definition "Information used to configure the services, parameters, and initial settings for a virtual server instance running in a cloud service.." .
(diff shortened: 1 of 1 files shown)


aamedina (commented):
Looks good to me :) 

I suggest editing the definition to be a bit more general and less specific as to terms like SaaS etc. 

I think a CloudService class for a second contribution makes sense. Perhaps CloudStorage would be a CloudService. And then the restriction would be that this Cloud Service Provider is a provider of some Cloud Service.

aamedina (commented):
Looks good to me

PR #329, opened 2024-12-13
Added Upper Model of OT Artifacts

OG3x100:


src/ontology/d3fend-protege.ttl
```diff
@@ -16121,6 +16126,97 @@ Operating System Monitoring Techniques have varied implementations including bui
             owl:onProperty :invokes ;
             owl:someValuesFrom :WriteMemory ] .
 
+:OTActuationDevice a owl:Class ;
+    rdfs:label "OT Actuation Device" ;
+    :definition "An OT Actuation device is a device capable of TCP/IP communication, has firmware and controls an OT Actuator." ;
```
netfl0, line 16131: remote TCP/IP from definition

netfl0:
@aamedina can you give this a review / check?

src/ontology/d3fend-protege.ttl
```diff
@@ -18808,7 +18886,7 @@ Legitimate emails from a sender may receive a lower trust rating over time if th
     rdfs:label "Sensor" ;
     rdfs:subClassOf :D3FENDCore,
         :DigitalInformationBearer ;
-    :definition "In the broadest definition, a sensor is a device, module, machine, or subsystem that detects events or changes in its environment and sends the information to other electronics, frequently a computer processor." ;
+    :definition "In the broadest definition, a sensor is a device, module, machine, or subsystem that detects events or changes in its environment and sends the information to other electronics, frequently a processor or logic solver." ;
```
netfl0, line 18889: This is not good, please remove.

src/ontology/d3fend-protege.ttl
```diff
@@ -16249,6 +16264,69 @@ Operating System Monitoring Techniques have varied implementations including bui
...
+            owl:onProperty :communicates-with ;
+            owl:someValuesFrom :OTSensor ] ;
+    rdfs:comment "Rockwell Compact 5000 IO Module",
+        "There are many types of I/O modules, including: analog input, analog output, HART input, HART output, digital input, digital output, mV input, pulse input, universal I/O, vibration input, and many other types of input or output modules.The functionality of the I/O Module can be embedded in the controller or as a separate module connected via chassis or I/O link." ;
+    rdfs:isDefinedBy <https://consteel-electronics.com/articles/what-is-IO-module> ;
+    :definition "An I/O module is a hardware component designed to connect to external devices and sensors, converting analog or digital signals into a format that can be processed by a control system and vice versa." ;
+    rdfs:seeAlso <https://www.rockwellautomation.com/en-us/support/documentation/technical/i-o/compact-5000-i-o-modules.html> ;
+    :synonym "IO Module" .
+
+:OTPowerSupply a owl:Class ;
```
aamedina, line 16311: I think we need a d3f:PowerSupply artifact that an OT Power Supply is a specialization of.

src/ontology/d3fend-protege.ttl
```diff
@@ -168,6 +168,11 @@
     rdfs:label "claims" ;
     rdfs:subPropertyOf :d3fend-catalog-object-property .
 
+:communicates-with a owl:ObjectProperty ;
```
aamedina, line 171: I think this is a owl:SymmetricProperty, where if x communicates with y then y also communicates with x. Otherwise, perhaps the d3f:connects property is sufficient.

src/ontology/d3fend-protege.ttl
```diff
@@ -203,6 +208,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :d3fend-catalog-object-property ;
     rdfs:range owl:Thing .
 
+:controls a owl:ObjectProperty ;
```
aamedina, line 211: I think ideally this would be a superproperty of d3f:manages, but I can see why you didn't choose that as the definition there is too specific to relating some thing x to a digital artifact y. 

d3f:manages rdfs:subPropertyOf d3f:controls makes sense to me. 

I think we should also add the inverse property (d3f:controlled-by).

src/ontology/d3fend-protege.ttl
```diff
@@ -957,6 +967,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:isDefinedBy <http://purl.obolibrary.org/obo/BFO_0000056> ;
     :definition "x participates-in y: The object x takes part in the event y, signifying that x contributes to or is affected by the event’s occurrence in some way." .
 
+:powered-by a owl:ObjectProperty ;
```
aamedina, line 970: I think this is a subproperty of d3f:dependsOn, and also needs an inverse (d3f:powers).

aamedina (changes requested):

PR #354, opened 2025-02-11
Attacker taxonomy

netfl0:


src/ontology/d3fend-protege.ttl
```diff
@@ -19688,6 +19731,17 @@ Parsons, L., Haque, E., & Liu, H. (2004). Subspace Clustering for High Dimension
...
 
+:SupplyChainAttacker a owl:Class ;
+    rdfs:label "Supply Chain Attacker" ;
+    rdfs:subClassOf :Attacker,
+        [ a owl:Restriction ;
+            owl:onProperty :accesses ;
+            owl:someValuesFrom :Software ],
+        [ a owl:Restriction ;
+            owl:onProperty :accesses ;
+            owl:someValuesFrom :T1592.001 ] ;
```
aamedina, line 19742: It’s not clear what it means for an attacker to 'access' an offensive technique. Typically, offensive techniques—like T1592.001, which involves gathering information about the victims hardware, are methods or procedures that an attacker employs rather than something to be directly accessed.

aamedina: I'd also like to note that a supply chain attacker is focused on targeting the dependencies of software or hardware. So they're not necessarily accessing software, but compromising the constituent components or dependencies, which then impacts the overall security. Maybe d3f:deceives-with :Software or :Hardware. I think we probably need a better semantic way to express this.

aamedina (changes requested):
this is really cool, my main feedback is that I am not sure if the supply chain attacker semantics make sense. 

Also, could we express that an Attacker employs offensive techniques with the goal to access? Does a d3f:targets semantic (which implies the attacker has the goal to exploit some artifact) make sense?

src/ontology/d3fend-protege.ttl
giacomodecolle: The current definition of "accesses" that I can read is "x accesses y: An subject x takes the action of reading from, writing into, or executing the stored information in the object y. Reads, writes, and executes are specific cases of accesses."

Agreed with @aamedina I'd rather say that techniques are akin to plans that direct or cco:prescribe certain actions. An attacker then would follow techniques, or implement techniques. Perhaps an attacker d3f:uses a technique?

lzrenterprises: Good catch, this was a bug! I don't even know how this got in there. It makes no sense 😂

netfl0: You guys are generous that was a spurious entry on my part. Not even sure how that happened! Will remove it.

src/ontology/d3fend-protege.ttl
```diff
@@ -2448,6 +2448,11 @@ Day, O., & Khoshgoftaar, T.M. (2017). A survey on heterogeneous transfer learnin
     rdfs:label "ATTACK Enterprise Thing" ;
     rdfs:subClassOf :ATTACKThing .
 
+:Attacker a owl:Class ;
+    rdfs:label "Attacker" ;
+    rdfs:subClassOf :Agent ;
+    :definition "An agent that attempts to exploit vulnerabilities to gain unauthorized access to data or systems." .
```
giacomodecolle, line 2454: Are all attacks requiring unauthorized access? For example, can I delete or damage or encrypt data without gaining unauthorized access to them?

src/ontology/d3fend-protege.ttl
```diff
@@ -27427,6 +27481,19 @@ Transformer-XL. (n.d.). Papers with Code. [Link](https://paperswithcode.com/meth
...
+    rdfs:subClassOf :TrustedAttacker,
+        [ a owl:Restriction ;
+            owl:onProperty :accesses ;
+            owl:someValuesFrom :PrivilegedUserAccount ] ;
+    :definition "A trusted attacker who misuses administrative access to execute attacks, often with elevated privileges." .
+
+:TrustedAttacker a owl:Class ;
+    rdfs:label "Trusted Attacker" ;
+    rdfs:subClassOf :Attacker ;
+    :definition "An individual within the organization who exploits their access to conduct unauthorized actions, either intentionally or through negligence." .
```
giacomodecolle, line 27495: I am splitting hairs here, but what about an external party which is trusted by the organization but is not a part of it? Say, a contractor or a guest

src/ontology/d3fend-protege.ttl
netfl0: I'd argue that your example was a form of unauthorized access.

Perhaps "unintended" ?

Open to a better suggestion here.

giacomodecolle: I guess this depends on what we mean by "access" rather than "authorized". Am I accessing a computer by being close to it and being able to destroy it, even if I can't use it or log in? Same for :PhysicalAttacker which mentions "physical access to devices"

netfl0:
> this is really cool, my main feedback is that I am not sure if the supply chain attacker semantics make sense.
> 
> Also, could we express that an Attacker employs offensive techniques with the goal to access? Does a d3f:targets semantic (which implies the attacker has the goal to exploit some artifact) make sense?

Agree semantics could improve quite a bit. We may want to redesign this in the very near future along the lines of your recommendation. Merging it to get people thinking about this.

Issue #368, opened 2025-03-07
Add Risk to Offensive Actions


bbreisch26:
d3f:OffensiveAction needs to capture risk via impact and likelihood data properties.

Want to add d3f:impact and d3f:likelihood

aamedina:
fixed in #369

PR #369, opened 2025-03-07
Add Risk to Offensive Action

bbreisch26:
#368 

Adds:
d3f:has-impact-rating
d3f:has-likelihood-rating

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -15578,13 +15578,7 @@ Programmatically checking if a pointer is NULL before use.
             owl:someValuesFrom :OffensiveTactic ],
         [ a owl:Restriction ;
             owl:onProperty :initiates ;
-            owl:someValuesFrom :Access ],
-        [ a owl:Restriction ;
-            owl:onProperty :has-impact-rating ;
-            owl:someValuesFrom :has-impact-rating ],
-        [ a owl:Restriction ;
-            owl:onProperty :has-likelihood-rating ;
-            owl:someValuesFrom :has-impact-rating ] ;
+            owl:someValuesFrom :Access ] ;
     rdfs:isDefinedBy <https://attack.mitre.org/docs/ATTACK_Design_and_Philosophy_March_2020.pdf> ;
     :display-baseurl "/offensive-technique/attack/" .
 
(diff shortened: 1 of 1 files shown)


aamedina (changes requested):
Great job with the datatype properties themselves.

I don't think Offensive Action is the right place to put these though. We don't have any Offensive Actions defined in the ontology yet. I think users would be expecting to assess the risk of offensive techniques. 

Also, with the rdfs:range declared you don't need to repeat it in the someValuesFrom. 

To me, it feels like these properties should be defined on anything considered a "Risk".

hack-sentinel:
We'll have to be careful with definitions for these... and we will need definitions or just be adding to confusion/ambiguity if they get used different ways in different models.   Both risk-impact and risk-likelihood will need to clearly defined as, say more of a risk cube sense (and whose definitions of those) or probabilistic (and probability vs likelihood is whole separate and statistically arcane arena), or whatever sense we want to mean them.

hack-sentinel (changes requested):
The most relevant/canonical reference to this scheme is here https://www.cto.mil/wp-content/uploads/2024/05/RIO-2023-2-2.pdf, which is found here https://www.cto.mil/sea/pg/

The language in the DoD document is not "impact" but "consequence".  If we keep "risk-impact", we at least need to mention in d3f:definition that "impact" is just a synonym for "consequence".  

To cite by title (not handle other than filename that I can see would be "Department of Defense Risk, Issue, and Opportunity (RIO) Management Guide for Defense Acquisition Programs".  That's awkwardly long for a definition, so let's add an isDefinedBy for each property and point it to the link on the document.

This will help folks know whether this is the property they want and whether what they're doing with it matches (near enough) the 5x5 matrix and intent.

Curiously, https://csrc.nist.gov/pubs/sp/800/30/r1/final (NIST RMF-related and similar linear scales) does not include an actual matrix or mention of a matrix as far as I could tell. Though it has been cited by others as origination of cyber risk matrices!

bbreisch26:
- renamed to risk-impact/risk-likelihood
- expanded definition of impact to note relationship to consequence
- Added isDefinedBy links (thanks @hack-sentinel !)

Awaiting further discussion between @netfl0 and @aamedina to decide on whether this truly applies to d3f:OffensiveAction.

netfl0:
After much discussion we will support two methods for capturing Risk concepts in a D3FEND Graph.

First, a simple method where we consider the offensive action to be a "risk event" as defined in [RIO](https://www.cto.mil/wp-content/uploads/2024/05/RIO-2023-2-2.pdf).

To capture the notion of a risk register "row" or "d3f:DatabaseRecord", more complex modeling is support in D3FEND CAD where you can reify all columns of the row as separate elements, e.g. "Risk Holder" as a relationship and object property to a custom node element. Documentation to come (for CAD).

PR #380, opened 2025-03-24
add initial precedence relationships between digital events

### LINKED ISSUE #379: Initial stab at "precedes" relationships between events.
We need an initial cut of these relationships (restrictions) added.

Criteria:
`precedes` and `preceded-by` relationships should only be added when it is **always** true.

E.g.

Image

A response should always be preceded-by a request.

Alternatively, a request does not always result in a response (many reasons might occur).

We will consider "maybe precedes"  later.

The UI on CAD needs to be improved to take advantage of this, that will be on a separate ticket.

aamedina:
fixes #379

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -1877,7 +1877,13 @@ GeeksforGeeks. (n.d.). Adaptive Resonance Theory (ART). [Link](https://www.geeks
 
 :AddUserToGroupEvent a owl:Class ;
     rdfs:label "Add User to Group Event" ;
-    rdfs:subClassOf :GroupManagementEvent ;
+    rdfs:subClassOf :GroupManagementEvent,
+        [ a owl:Restriction ;
+            owl:onProperty :preceded-by ;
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -2213,7 +2229,8 @@ Hardening an application's configuration involves analyzing not only the applica
 Exploits may, for example, rely on knowledge of addresses in a process's memory, they may alter memory contents, and they may cause a program to use instructions in a way that they were not intended.  By, for example, including code that dynamically changes the memory address of data or code on each run, introducing logic to validating the memory contents before certain potentially dangerous flows are executed, or monitoring a program for unusual sequence of instructions, this makes it harder for an attacker to craft a working exploit.""" ;
     :synonym "Process Hardening" .
 
-:ApplicationInstallationEvent a owl:Class ;
+:ApplicationInstallationEvent a owl:Class,
+        owl:NamedIndividual ;
```
netfl0, line 2233: delete NI

PR #381, opened 2025-03-27
CDR Techniques

ryantxu1:


src/ontology/d3fend-protege.ttl
```diff
@@ -9189,6 +9384,12 @@ Esri. (n.d.). How CycleGAN Works. [Link](https://developers.arcgis.com/python/gu
 :D3FENDKBThing a owl:Class ;
     rdfs:label "D3FEND KB Thing" .
 
+:Data a owl:Class ;
```
netfl0, line 9387: This should just be a synonym for DigitalInformation.

d3f:DigitalInformation is data :)

PR #382, opened 2025-04-01
Add missing tactic to CSV cleaning script.

rmaxted-wowcorp:
We rely on the output of this script (i.e. https://d3fend.mitre.org/ontologies/d3fend.csv) and found it to be missing the Restore tactic.

This PR should resolve this issue.


netfl0:
Thank you!

Issue #372, opened 2025-03-10
Access Modeling needs to include identity


vikhari:
Also, IAM architecture often maps the PAP (Policy Administration Point), PDP (Policy Decision Point), PIP (Policy Information Point), and PEP (Policy Enforcement Point) principles. This involves understanding not just identifying and recording the access permissions granted but the manner in which identity information and identities are transferred between components. Although the information could be categorized as a Data Exchange Mapping the notion is more Access modeling (identity/authN/authZ terminology). Should Access Modeling language be slightly tweaked to include identity?

vikhari:
PR is #400

PR #400, opened 2025-04-26
Added 'identity'  to the definition of D3-AM (Issue #372)

vikhari:


netfl0:
@hack-sentinel please review. I agree with some of the proposed changes but not all (e.g. `enumerates`).

PR #421, opened 2025-06-18
Enhancing Artifact Relationships Pt 1

ryantxu1:
Mostly focusing on Digital Information Artifacts

Relationships Added:

d3f:Command
	- Definition changed

(NEW) d3f:ShellCommand
	- subclass of d3f:Command
	- recorded-in some d3f:CommandHistoryLog
	- may-create some d3f:Process
	- may-execute some d3f:ExecutableFile
	
d3f:RemoteCommand
	- produces some d3f:NetworkTraffic

d3f:DatabaseQuery
	- queries some d3f:Database 

d3f:CryptographicKey
	- associated-with some Certificate
	
d3f:UserInterface
	- interprets some d3f:Command

d3f:PrivateKey
	- associated-with some d3f:PublicKey

d3f:PublicKey
	- associated-with some d3f:PrivateKey
	- contained-by some d3f:Certificate

d3f:PasswordManager
	- contains some d3f:Credential

d3f:ContainerOrchestrationSoftware
	- manages some d3f:ContainerProcess

d3f:CredentialManagementSystem
  - manages some d3f:Credential

d3f:WebServerApplication
	- contained-by some d3f:WebServer

d3f:WebServer
 	- may-contain some d3f:WebServerApplication

d3f:SoftwarePatch
	- updates some d3f:Software

d3f:DynamicAnalysis
	- uses some d3f:DynamicAnalysisTool

d3f:Firmware
	- controls some d3f:EmbeddedComputer
	- NOT SURE ON THIS ONE

d3f:GraphicsCardFirmware
	- contained-by some d3f:GraphicsProcessingUnit

d3f:GraphicsProcessingUnit
	- contains some d3f:GraphicsCardFirmware

d3f:HardDiskFirmware
	- contained-by some d3f:SecondaryStorage

d3f:SecondaryStorage
	- may-contain some d3f:HardDiskFirmware

d3f:NetworkCardFirmware
	- contained-by some d3f:NetworkInterfaceCard

d3f:NetworkInterfaceCard
	- contains some d3f:NetworkCardFirmware

d3f:CommandLineInterface
	- contains some d3f:ConsoleOutputFunction

d3f:ExternalContentInclusionFunction
	- modifies some d3f:File

d3f:LogMessageFunction
	- recorded-in some d3f:Log

d3f:StoredProcedure
	- associated-with some d3f:Database 

d3f:LocalAuthenticationService
	- authenticates some d3f:LocalUserAccount

d3f:LocalAuthorizationService
	- authorizes some d3f:LocalUserAccount

d3f:Host-basedFirweall
	- subclass of Firewall

d3f:StoredProcedure
	- contained-by some Database


aamedina (changes requested):
minor changes requested, thanks!

src/ontology/d3fend-protege.ttl
```diff
@@ -3037,7 +3037,10 @@ Sciencedirect. (n.d.). Semi-supervised learning: An overview. [Link](https://www
 :BootLoader a owl:Class ;
     rdfs:label "Boot Loader" ;
     skos:altLabel "Bootloader" ;
-    rdfs:subClassOf :Software ;
+    rdfs:subClassOf :Software,
+        [ a owl:Restriction ;
+            owl:onProperty :instructs ;
+            owl:someValuesFrom :SystemInitProcess ] ;
```
netfl0, line 3043: This hands over control to the kernel I thought, then init happens afterwards to start services (ala init.d). 

 Presumably there is diversity here, but i want to avoid confusion with the init.d sense of init.

src/ontology/d3fend-protege.ttl
```diff
@@ -3869,7 +3872,16 @@ Wikipedia. (n.d.). Coefficient of variation. [Link](https://en.wikipedia.org/wik
 
 :Command a owl:Class ;
     rdfs:label "Command" ;
-    rdfs:subClassOf :DigitalInformation ;
+    rdfs:subClassOf :DigitalInformation,
```
netfl0, line 3875: I think we need to make :Command a bit more general. The current definition of command needs to be fixed, it describes Shell Commands more so than general commands. I'd put these restrictions on Shell commands since they might create processes.

src/ontology/d3fend-protege.ttl
```diff
@@ -4090,7 +4102,10 @@ Another approach passively inspects network traffic with application protocol an
 
 :ConsoleOutputFunction a owl:Class ;
     rdfs:label "Console Output Function" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :associated-with ;
```
netfl0, line 4107: we don't want to use associated-with like this.

I think you can say CommandLineInterface contains :ConsoleOutputFunction

We don't need to go the other way.

src/ontology/d3fend-protege.ttl
```diff
@@ -22652,7 +22730,10 @@ Application-layer discovery:
 :SoftwarePatch a owl:Class ;
     rdfs:label "Software Patch" ;
     skos:altLabel "Patch" ;
-    rdfs:subClassOf :Software ;
+    rdfs:subClassOf :Software,
```
netfl0, line 22733: true but please remove

src/ontology/d3fend-protege.ttl
```diff
@@ -21911,7 +21971,10 @@ List of known unauthorized script files or regular expression patterns must be k
 :SecondaryStorage a owl:Class ;
     rdfs:label "Secondary Storage" ;
     rdfs:subClassOf :HardwareDevice,
-        :Storage ;
+        :Storage,
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :HardDiskFirmware ] ;
```
netfl0, line 21977: please remove.

src/ontology/d3fend-protege.ttl
```diff
@@ -21270,7 +21327,10 @@ Complex intranet VPNs or routing encapsulation may affect the detection analytic
 
 :RemoteCommand a owl:Class ;
     rdfs:label "Remote Command" ;
-    rdfs:subClassOf :Command ;
+    rdfs:subClassOf :Command,
+        [ a owl:Restriction ;
+            owl:onProperty :produces ;
+            owl:someValuesFrom :NetworkTraffic ] ;
```
netfl0, line 21333: the command is contained by the network traffic, it does not produce it per se.

The thing that sends commands would be producing it.

src/ontology/d3fend-protege.ttl
```diff
@@ -13538,7 +13552,10 @@ File format specifications often define expected values for specific fields. A c
 
 :Firmware a owl:Class ;
     rdfs:label "Firmware" ;
-    rdfs:subClassOf :Software ;
+    rdfs:subClassOf :Software,
+        [ a owl:Restriction ;
+            owl:onProperty :controls ;
+            owl:someValuesFrom :EmbeddedComputer ] ;
```
hack-sentinel, line 13558: 1. Firmware provides instructions for more than embedded computers, could be any Computer Platform. If we want this relationship to specifically link to EmbededComputer we'd need an EmbeddedComputerFirmware subclass of Firmware.
2. The property instructs is a better match for this than controls and would fit current definition.

src/ontology/d3fend-protege.ttl
```diff
@@ -14043,7 +14066,10 @@ Talukdar, P. (2020, June 10). Convolutional Neural Networks Explained. Towards D
 :HardDiskFirmware a owl:Class ;
     rdfs:label "Hard Disk Firmware" ;
     skos:altLabel "Hard Drive Firmware" ;
-    rdfs:subClassOf :PeripheralFirmware ;
+    rdfs:subClassOf :PeripheralFirmware,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
+            owl:someValuesFrom :SecondaryStorage ] ;
```
hack-sentinel, line 14072: We should add HardDrive as a class, subclass of SecondaryStorage to allow parallel formation with all the sibling device firmwares here which each have corresponding devices and then have contained-by.  Not sure why we don't have HardDrive already. [Next round we can add SSD and SSDFirmware.]

src/ontology/d3fend-protege.ttl
```diff
@@ -14389,7 +14415,8 @@ A homoglyph, in this context, is a deceptive string or word which looks like a t
 
 :Host-basedFirewall a owl:Class ;
     rdfs:label "Host-based Firewall" ;
-    rdfs:subClassOf :SystemSoftware ;
+    rdfs:subClassOf :Firewall,
```
hack-sentinel, line 14418: As it stands, making Host-basedFirewall a subclass of Firewall will make Host-baseFirewall a Network Node (and thus a Computing Platform.  Drop Firewall here for this pass.

We would need to define a new Firewall concept to subclass as above and move the Firewall we already have to Network Firewall Device (or some such name).  Here the firewall is part of the node and implemented by software on the node, rather being a classic firewall device/node.  For this pass, we should drop Firewall or it will get weird and make an issue to factor out the Firewall capability/function from the node per se.

src/ontology/d3fend-protege.ttl
```diff
@@ -16193,13 +16220,19 @@ Newer system call.""" .
 :LocalAuthenticationService a owl:Class ;
     rdfs:label "Local Authentication Service" ;
     rdfs:subClassOf :AuthenticationService,
-        :SystemServiceSoftware ;
+        :SystemServiceSoftware,
+        [ a owl:Restriction ;
+            owl:onProperty :authenticates ;
+            owl:someValuesFrom :LocalUserAccount ] ;
```
hack-sentinel, line 16226: I would drop the subclass SystemServiceSofware here.  The service is currently modeled as a process (that would be be instructed by the software).

src/ontology/d3fend-protege.ttl
```diff
@@ -16193,13 +16220,19 @@ Newer system call.""" .
...
     :definition "A local authentication service running on a host can authenticate a user logged into just that local host computer." .
 
 :LocalAuthorizationService a owl:Class ;
     rdfs:label "Local Authorization Service" ;
     rdfs:subClassOf :AuthorizationService,
-        :SystemServiceSoftware ;
+        :SystemServiceSoftware,
```
hack-sentinel, line 16232: Drop superclass of SystemServiceSoftware.

src/ontology/d3fend-protege.ttl
```diff
@@ -16364,7 +16397,10 @@ Wikipedia. (n.d.). Logistic regression. [Link](https://en.wikipedia.org/wiki/Log
 
 :LogMessageFunction a owl:Class ;
     rdfs:label "Log Message Function" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :recorded-in ;
```
hack-sentinel, line 16402: The function modifies the Log.  The relation recorded-in is the relation between an event and the record (or log) it is stored.  In short, we don't "record" the function.  LogMessageFunction records DigitalEventRecord would be valid though, so that could be added as well, or even instead of LogMessageFunction modifies some Log.

src/ontology/d3fend-protege.ttl
```diff
@@ -21529,7 +21583,10 @@ Complex intranet VPNs or routing encapsulation may affect the detection analytic
 
 :RemoteCommand a owl:Class ;
     rdfs:label "Remote Command" ;
-    rdfs:subClassOf :Command ;
+    rdfs:subClassOf :Command,
+        [ a owl:Restriction ;
+            owl:onProperty :produces ;
+            owl:someValuesFrom :NetworkTraffic ] ;
```
hack-sentinel, line 21589: produces here is close but Peter may have had a thought on this I don't see here.  double-check for Peter's comment on prior delta here.

src/ontology/d3fend-protege.ttl
```diff
@@ -22174,7 +22231,10 @@ List of known unauthorized script files or regular expression patterns must be k
 :SecondaryStorage a owl:Class ;
     rdfs:label "Secondary Storage" ;
     rdfs:subClassOf :HardwareDevice,
-        :Storage ;
+        :Storage,
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :HardDiskFirmware ] ;
```
hack-sentinel, line 22237: This is definitely the counterpart to the inverse restriction indicating need for HardDisk add as subclass of SecondaryStorage. Not all SecondaryStorage is a HardDrive (i.e, magnetic disk).

src/ontology/d3fend-protege.ttl
```diff
@@ -31699,7 +31780,10 @@ Geolocation data for each user logon attempt is collected and used to create a b
 :UserInterface a owl:Class ;
     rdfs:label "User Interface" ;
     skos:altLabel "UI" ;
-    rdfs:subClassOf :DigitalInformationBearer ;
+    rdfs:subClassOf :DigitalInformationBearer,
+        [ a owl:Restriction ;
+            owl:onProperty :interprets ;
+            owl:someValuesFrom :ShellCommand ] ;
```
hack-sentinel, line 31786: Not all UserInterfaces are canonically able to intepret ShellCommands.  Better to tie this restriction to CommandLineInterface for the definite property "interprets". If we want to capture that some GUIs, with extensions, can interpret commands through some softer or through terminal or shell within overall GUI system, then just add may-interpret (we already have the property!) as a second restriction from Graphical User Interface.  We don't need to break out Shell or Terminal just yet to get some value here.  But Shell might be something to consider in future.

src/ontology/d3fend-protege.ttl
```diff
@@ -32045,17 +32129,23 @@ An ID token is proof of the user's authentication. An ID token is encoded as a J
 
 :WebServer a owl:Class ;
     rdfs:label "Web Server" ;
-    rdfs:subClassOf :Server ;
+    rdfs:subClassOf :Server,
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :WebServerApplication ] ;
```
hack-sentinel, line 32135: We should first rename WebServerApplication to WebApplication, which is the better name for the definition we are using for WebServerApplication.  

We then should create a new (correct) WebServerApplication ask Grok, Google, or OpenAI what a "web server application" is <sigh> (could alias or name that to be WebServerSoftware) [https://en.wikipedia.org/wiki/Comparison_of_web_server_software class] for things like Apache, nginx.

With those distinctions in hand:  WebServer contains WebServerApplication; may-contain WebApplication.

hack-sentinel: Minimum should be to rename WebServerApplication->WebApplication, as the former will confuse many a folk.

src/ontology/d3fend-protege.ttl
```diff
@@ -32045,17 +32129,23 @@ An ID token is proof of the user's authentication. An ID token is encoded as a J
...
 
 :WebServerApplication a owl:Class ;
     rdfs:label "Web Server Application" ;
     skos:altLabel "Web App",
         "Web Application" ;
-    rdfs:subClassOf :ServiceApplication ;
+    rdfs:subClassOf :ServiceApplication,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
+            owl:someValuesFrom :WebServer ] ;
```
hack-sentinel, line 32146: Correct language here in definition and correct inverse relations to match work above.  Somehow we confused and did not catch using the term web server application instead of simply web application. <sigh>

hack-sentinel (changes requested):
Requesting changes.

ryantxu1:
@ikiril01 per Mike, taking out that GPU/firmware relationship for now (this release). Made #445 to track in the near future

src/ontology/d3fend-protege.ttl
```diff
@@ -3915,8 +3915,7 @@ Wikipedia. (n.d.). Coefficient of variation. [Link](https://en.wikipedia.org/wik
 :Command a owl:Class ;
     rdfs:label "Command" ;
     rdfs:subClassOf :DigitalInformation ;
-    rdfs:isDefinedBy <http://dbpedia.org/resource/Command_(computing)> ;
-    :definition "In computing, a command is a directive to a computer program acting as an interpreter of some kind, in order to perform a specific task. Most commonly a command is either a directive to some kind of command-line interface, such as a shell, or an event in a graphical user interface triggered by the user selecting an option in a menu." .
+    :definition "A directive or instruction issued to a computer system, software, or hardware component to perform a specific action, operation, or computation." .
```
netfl0, line 3918: update to:

A directive (i.e., an instruction specifying a procedure) which, when issued to a computer system, software, or hardware component, causes that entity to execute a specific action, operation, or computation.

Also keep the isDefby from before

src/ontology/d3fend-protege.ttl
```diff
@@ -12263,7 +12268,10 @@ CompNeuro Neuromatch Academy Tutorials. [Link](https://compneuro.neuromatch.io/t
             owl:someValuesFrom :DocumentFile ],
         [ a owl:Restriction ;
             owl:onProperty :analyzes ;
-            owl:someValuesFrom :ExecutableFile ] ;
+            owl:someValuesFrom :ExecutableFile ],
+        [ a owl:Restriction ;
```
netfl0, line 12272: remove this.

src/ontology/d3fend-protege.ttl
```diff
@@ -12899,7 +12907,10 @@ Intro to Active Learning. inovex Blog. [Link](https://www.inovex.de/de/blog/intr
 
 :ExternalContentInclusionFunction a owl:Class ;
     rdfs:label "External Content Inclusion Function" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :modifies ;
+            owl:someValuesFrom :File ] ;
     :definition "External content, strings or data, are inserted into a local document (e.g. xml document) as if it were a native part of that document." .
```
netfl0, line 12914: change definition to: 

A subroutine which handles a content inclusion directive from an original file. When invoked, the external content is included in the resulting open file.

src/ontology/d3fend-protege.ttl
```diff
@@ -12899,7 +12907,10 @@ Intro to Active Learning. inovex Blog. [Link](https://www.inovex.de/de/blog/intr
 
 :ExternalContentInclusionFunction a owl:Class ;
     rdfs:label "External Content Inclusion Function" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :modifies ;
```
netfl0: should be d3f:accesses

src/ontology/d3fend-protege.ttl
```diff
@@ -13972,7 +13983,10 @@ Yang, S., Pan, L., & Cheng, J. (2021). Graph-based Semi-Supervised Learning Meth
 :GraphicsCardFirmware a owl:Class ;
     rdfs:label "Graphics Card Firmware" ;
     skos:altLabel "Video Card Firmware" ;
-    rdfs:subClassOf :PeripheralFirmware ;
+    rdfs:subClassOf :PeripheralFirmware,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
```
netfl0: reverse this and delete contained-by

src/ontology/d3fend-protege.ttl
```diff
@@ -16364,7 +16382,10 @@ Wikipedia. (n.d.). Logistic regression. [Link](https://en.wikipedia.org/wiki/Log
 
 :LogMessageFunction a owl:Class ;
     rdfs:label "Log Message Function" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :modifies ;
```
netfl0: d3f:produces d3f:DigitalEventRecord

src/ontology/d3fend-protege.ttl
```diff
@@ -17090,7 +17111,10 @@ Network Access Mediation is a crucial process in telecommunications and IT netwo
 :NetworkCardFirmware a owl:Class ;
     rdfs:label "Network Card Firmware" ;
     skos:altLabel "Network Controller Firmware" ;
-    rdfs:subClassOf :PeripheralFirmware ;
+    rdfs:subClassOf :PeripheralFirmware,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
```
netfl0: fix the same way as GPU firmware

src/ontology/d3fend-protege.ttl
```diff
@@ -21202,7 +21235,13 @@ Proxy-based Web Server Access Mediation involves controlling access to web serve
 
 :PublicKey a owl:Class ;
     rdfs:label "Public Key" ;
-    rdfs:subClassOf :AsymmetricKey ;
+    rdfs:subClassOf :AsymmetricKey,
+        [ a owl:Restriction ;
+            owl:onProperty :associated-with ;
```
netfl0: d3f:depends-on Private key

src/ontology/d3fend-protege.ttl
```diff
@@ -21202,7 +21235,13 @@ Proxy-based Web Server Access Mediation involves controlling access to web serve
...
+            owl:someValuesFrom :PrivateKey ],
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
```
netfl0: reverse the inverse and delete

src/ontology/d3fend-protege.ttl
```diff
@@ -20513,7 +20543,10 @@ Wikipedia. (n.d.). Principal component analysis. [Link](https://en.wikipedia.org
 
 :PrivateKey a owl:Class ;
     rdfs:label "Private Key" ;
-    rdfs:subClassOf :AsymmetricKey ;
+    rdfs:subClassOf :AsymmetricKey,
+        [ a owl:Restriction ;
+            owl:onProperty :associated-with ;
```
netfl0: d3f:has-dependent

src/ontology/d3fend-protege.ttl
```diff
@@ -21529,7 +21568,10 @@ Complex intranet VPNs or routing encapsulation may affect the detection analytic
 
 :RemoteCommand a owl:Class ;
     rdfs:label "Remote Command" ;
-    rdfs:subClassOf :Command ;
+    rdfs:subClassOf :Command,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
```
netfl0: reverse this but should be d3f:may-contain

src/ontology/d3fend-protege.ttl
```diff
@@ -23276,7 +23333,10 @@ Wolfram MathWorld. (n.d.). Statistics. [Link](https://mathworld.wolfram.com/Stat
 
 :StoredProcedure a owl:Class ;
     rdfs:label "Stored Procedure" ;
-    rdfs:subClassOf :Subroutine ;
+    rdfs:subClassOf :Subroutine,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
```
netfl0: reverse please with may-contain

src/ontology/d3fend-protege.ttl
```diff
@@ -32049,14 +32123,6 @@ An ID token is proof of the user's authentication. An ID token is encoded as a J
     rdfs:isDefinedBy <http://dbpedia.org/resource/Web_server> ;
     :definition "A web server is server software, or hardware dedicated to running this software, that can satisfy client requests on the World Wide Web. A web server can, in general, contain one or more websites. A web server processes incoming network requests over HTTP and several other related protocols. While the major function is to serve content, a full implementation of HTTP also includes ways of receiving content from clients. This feature is used for submitting web forms, including uploading of files." .
 
-:WebServerApplication a owl:Class ;
```
netfl0, line 34757: add this back, this would be "nginx"

Not a synonym for webapp...

definitions need improvement

    rdfs:isDefinedBy <http://dbpedia.org/resource/Web_application>  

is wrong, we can remove and or find another link.

src/ontology/d3fend-protege.ttl
```diff
@@ -31977,6 +32037,17 @@ It is a combination of VPN hardware and software technologies that provides VPN
...
+:WebApplication a owl:Class ;
+    rdfs:label "Web Application" ;
+    skos:altLabel "Web App",
+        "Web Application" ;
+    rdfs:subClassOf :ServiceApplication,
+        [ a owl:Restriction ;
+            owl:onProperty :contained-by ;
+            owl:someValuesFrom :WebApplicationServer ] ;
+    rdfs:isDefinedBy <http://dbpedia.org/resource/Web_application> ;
+    :definition "A web server application (or web app) is an application software that runs on a web server, unlike computer-based software programs that are stored locally on the Operating System (OS) of the device. Web applications are accessed by the user through a web browser with an active internet connection. These applications are programmed using a client-server modeled structure-the user (\"client\") is provided services through an off-site server that is hosted by a third-party. Examples of commonly-used web applications include: web-mail, online retail sales, online banking, and online auctions." .
```
netfl0: definition:

An application which is delivered by a web server over HTTP protocols that is presented to a client web browser.

netfl0 (changes requested):

netfl0 (changes requested):
please see requested changes.

PR #442, opened 2025-07-23
Add inbound internet web traffic artifact

ryantxu1:


hack-sentinel (changes requested):
To be completely parallel to outbound side (OutboundInternetEncryptedWebTraffic), we need to add a rdfs:subClassOf InboundInternetWebTraffic to InboundInternetEncryptedWebTraffic.  That seems to be the general pattern for network traffic, so we should continue to follow that here.

PR #448, opened 2025-08-04
Make D3FEND technique a concatenation of label.

rmaxted-wowcorp:
Most techniques are simply a concatenation of their label.

We had an infrequent job choke on these two technique since it doesn't follow that guideline.

i.e. 
https://d3fend.mitre.org/technique/d3f:EndpointBasedWebServerAccessMediation/
https://d3fend.mitre.org/technique/d3f:ProxyBasedWebServerAccessMediation/
vs
https://d3fend.mitre.org/technique/d3f:Certificate-basedAuthentication/
https://d3fend.mitre.org/technique/d3f:Hardware-basedWriteProtection/
https://d3fend.mitre.org/technique/d3f:Token-basedAuthentication/
https://d3fend.mitre.org/technique/d3f:Application-basedProcessIsolation/
https://d3fend.mitre.org/technique/d3f:Hardware-basedProcessIsolation/
https://d3fend.mitre.org/technique/d3f:Kernel-basedProcessIsolation/

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -2,5 +2,7 @@
 
 1. Defensive Techniques are punned as instances of their parent classes.
 1. CamelCase  [IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier), capitalize all words in the IRI fragment.
+1. Labels for D3FEND defined terms must be in harmony with their IRI fragments.
+    1. e.g. An IRI of `Proxy-basedWebServerAccessMediation` would have the label `Proxy-based Web Server Access Mediation` 
 1. Do not use bracket characters in IRIs.
 1. Do not use unicode characters in IRIs.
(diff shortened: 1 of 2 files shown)


aamedina:
I hear what you're saying, and I would technically have no problem with this if we all agree upon a convention that would be applied in D3FEND universally. But generally speaking the IRI of a concept has no direct relationship to its label; though they often overlap, the IRI is just a globally unique identifier for the concept. Many ontologies just have a pure TitleCaseConvention, like a Java class name with no dashes. I acknowledge that right now the ontology is inconsistent in its conventions in this regard. @netfl0 may have thoughts

netfl0:
Agree, the label should be in harmony with IRI fragment.

Great find @rmaxted-wowcorp .

aamedina:
@netfl0 Would this convention only apply to D3FEND techniques? ATT&CK techniques, CWEs, CCI/NIST controls have different labels from their IRIs.

netfl0:
> @netfl0 Would this convention only apply to D3FEND techniques? ATT&CK techniques, CWEs, CCI/NIST controls have different labels from their IRIs.

D3FEND defined terms yes (much more than just techniques)

aamedina:
I'm good with that! @rmaxted-wowcorp can you also update the readme with this convention documented? Eventually we will want to formalize an ontology guide as a separate document.

rmaxted-wowcorp:
Thanks @netfl0 and @aamedina, I've updated the CONVENTIONS.md to reflect this convention.

Please let me know if there are any other required changes.

PR #466, opened 2025-12-03
Operating Mode Refactoring

ikiril01:
- Updated definition of `OperatingMode` to be more generic (i.e., for things like safe mode in spacecraft)
- Added new `OTControllerOperatingMode` as sub-class of Operating Mode
- Updated `OperatingModeMonitoring` & `OperatingModeRestriction` to point to `OTControllerOperatingMode`
- Added new `has-operating-mode` object property and updated `OTController` to have a restriction which points to `OTControllerOperatingMode`
- Added restriction to `T0858` to point to `OTControllerOperatingMode`
- Updated all OT*Command classes which pointed to `OperatingMode` to now point to `OTControllerOperatingMode`


holtorama:
Looks good, for the new definition of :OperatingMode "An operating mode is a user selectable option ...." maybe change "user selectable" as the device may change it's own. Perhaps along the lines of "The operating mode designates the state, sometimes user selectable, ..." @ikiril01

PR #472, opened 2025-12-09
SDR/IA-0002

ikiril01:
- Added SDR class + related classes (e.g., `SDRRadioConfiguration`
- Added DSP Application class
- Added SDR waveform class (subclass of DSP Application)
- Added RTOS class (subclass of OS)
- Added mappings to IA-0002


src/ontology/d3fend-protege.ttl
```diff
@@ -27532,6 +27563,38 @@ Wikipedia. (n.d.). Skewness. [Link](https://en.wikipedia.org/wiki/Skewness)""" .
...
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :Transmitter ],
+        [ a owl:Restriction ;
+            owl:onProperty :runs ;
+            owl:someValuesFrom :RealTimeOperatingSystem ] ;
+    rdfs:isDefinedBy <http://dbpedia.org/resource/Software-defined_radio> ;
+    :definition "Software-defined radio (SDR) is a radio communication system where components that conventionally have been implemented in analog hardware (e.g. mixers, filters, amplifiers, modulators/demodulators, detectors, etc.) are instead implemented by means of software on a computer or embedded system." ;
+    skos:altLabel "SDR" .
+
+:SoftwareDefinedRadioWaveform a owl:Class ;
```
netfl0: This appears to the configuration of a DSP Application which is a specification or description of an actual waveform.

Needs a better reference as well, are there standardized waveform representations/description languages.

ikiril01: @netfl0 I updated the definition using language directly from the SCA 4.1 spec. SCA is a commonly used waveform specification language, especially for military and gov't systems. https://en.wikipedia.org/wiki/Software_Communications_Architecture

src/ontology/d3fend-protege.ttl
```diff
@@ -28055,6 +28160,20 @@ A standalone honeynet does not directly interact with the real enterprise enviro
 A standalone honeynet is a lower risk to deploy compared to connected or integrated honeynets due to its isolation from the enterprise network. However, this comes at cost in loss of fidelity and realism. Significant extra effort must be made in order to make the environment look realistic.""" ;
     :kb-reference :Reference-DynamicSelectionAndGenerationOfAVirtualCloneForDetonationOfSuspiciousContentWithinAHoneyNetwork_PaloAltoNetworksInc .
 
+:StandaloneSDR a owl:Class ;
```
netfl0: Should be named something like:

SDRDevice
SDRAppliance
SDRComputer

src/ontology/d3fend-protege.ttl
```diff
@@ -26822,6 +26865,68 @@ Software installed on the host system hooks into a scripting engine to intercept
 List of known unauthorized script files or regular expression patterns must be kept up to date to ensure detection of new threats.""" ;
     :kb-reference :Reference-DetectingScript-basedMalware_CrowdstrikeInc .
 
+:SDR a owl:Class ;
```
netfl0: this should match the rdfs:label per style guide.

src/ontology/d3fend-protege.ttl
```diff
@@ -26582,6 +26617,14 @@ Queries for reverse resolution requests (that is, requests where IP(s) are sent
     :kb-reference :Reference-PrivacyAndSecuritySystemsAndMethodsOfUse,
         :Reference-Technical_Specifications_for_Construction_and_Management_of_Sensitive_Compartmented_Information_Facilities .
 
+:RFStateChangeEvent a owl:Class ;
```
netfl0: This is not an SDR Event. It would be an RF Event which i do not think we model.

netfl0 (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: This is addressed in the latest commit.

src/ontology/d3fend-protege.ttl
ikiril01: Should be fixed in the latest commit.

src/ontology/d3fend-protege.ttl
ikiril01: I've renamed this event and updated its description for accuracy, but LMK if it's still problematic.

src/ontology/d3fend-protege.ttl
```diff
@@ -800,6 +800,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:range :Artifact ;
     owl:inverseOf :may-have-weakness .
 
+:may-block a owl:ObjectProperty ;
```
hack-sentinel, line 814: Add to declaration of :blocks this:

:blocks rdfs:subProperty :may-block

src/ontology/d3fend-protege.ttl
```diff
@@ -26822,6 +26865,68 @@ Software installed on the host system hooks into a scripting engine to intercept
 List of known unauthorized script files or regular expression patterns must be kept up to date to ensure detection of new threats.""" ;
     :kb-reference :Reference-DetectingScript-basedMalware_CrowdstrikeInc .
 
+:SDR a owl:Class ;
+    rdfs:label "Software-defined Radio" ;
+    rdfs:subClassOf :HardwareDevice,
+        [ a owl:Restriction ;
+            owl:onProperty :contains ;
```
hack-sentinel: Always or sometimes send/transmit-only SDR?  Seems like this should be may-contain.  Surely one of these is present some (tx union rx) but two mays seem correct here.

src/ontology/d3fend-protege.ttl
```diff
@@ -15890,34 +15902,6 @@ SafeSEH might be applied only to some executable files or modules, allowing an a
             owl:someValuesFrom :ExecutableBinary ] ;
     rdfs:seeAlso <https://dbpedia.org/page/Exec> .
 
-:ExecutableAllowlisting a :ExecutableAllowlisting,
```
hack-sentinel, line 16085: Roll back to ExecutableAllowlisting please.  Don't think we really going with SoftwareArtifactAllowListing and deleting this.  This would also impact d3fend-id choice and D3-EAL.

src/ontology/d3fend-protege.ttl
```diff
@@ -800,6 +800,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:range :Artifact ;
     owl:inverseOf :may-have-weakness .
 
+:may-block a owl:ObjectProperty ;
+    rdfs:label "may-block" ;
+    rdfs:subPropertyOf :may-be-associated-with ;
+    :definition "x may-add y: They entity x may block the thing y; that is, 'x blocks y' may be true." .
```
hack-sentinel: definition should start "x may-block y:..." (not "may-add").

src/ontology/d3fend-protege.ttl
```diff
@@ -17068,6 +17061,12 @@ The DNS lookup can be blocked by either dropping the network traffic with an inl
     :kb-reference :Reference-UseDNSPolicyForApplyingFiltersOnDNSQueries ;
     :synonym "Forward Resolution IP Blacklisting" .
 
+:FPGABitstream a owl:Class ;
+    rdfs:label "FPGA Bitstream" ;
+    rdfs:subClassOf :ApplicationConfigurationFile ;
+    :definition "A binary configuration file generated by synthesizing and placing-and-routing an HDL design, which is loaded into an Field-Programmable Gate Array (FPGA) to physically define its internal logic, interconnects, and I/O behavior. Rather than being executed by a processor, it programs the device itself." ;
```
hack-sentinel: Change to "into **_a_** Field-Programmable" (that is, 'an' -> 'a')

src/ontology/d3fend-protege.ttl
```diff
@@ -27535,6 +27555,137 @@ Wikipedia. (n.d.). Skewness. [Link](https://en.wikipedia.org/wiki/Skewness)""" .
...
+    rdfs:seeAlso <https://en.wikipedia.org/wiki/Software-defined_radio#RTL-SDR> ;
+    :synonym "Peripheral SDR" .
+
+:Software-definedRadioEvent a owl:Class ;
+    rdfs:label "Software-defined Radio Event" ;
+    rdfs:subClassOf :DigitalEvent,
+        [ a owl:Restriction ;
+            owl:onProperty :has-participant ;
+            owl:someValuesFrom :Software-definedRadio ] ;
+    :definition "An event involving an software-defined radio (SDR) device indicating that the SDR's lifecycle state, operational state, configuration, data-streaming status, timing/reference status, or fault condition has changed." ;
```
hack-sentinel: change to "..Involving **_a_** software-defined radio.."

src/ontology/d3fend-protege.ttl
```diff
@@ -25694,6 +25707,13 @@ Additional examples of equipment include couplers, attenuators, power dividers,
             owl:onProperty :reads ;
             owl:someValuesFrom :MemoryBlock ] .
 
+:RealTimeOperatingSystem a owl:Class ;
```
hack-sentinel: If we are doing Software-definedRadio, then this would be Real-timeOperatingSystem for consistency.

src/ontology/d3fend-protege.ttl
```diff
@@ -16134,7 +16118,16 @@ On a Windows machine the Windows Defender Application Control (WDAC) policy enfo
...
+            owl:onProperty :may-modify ;
+            owl:someValuesFrom :Software-definedRadioConfiguration ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-modify ;
+            owl:someValuesFrom :Software-definedRadioWaveformApplication ],
+        [ a owl:Restriction ;
+            owl:onProperty :modifies ;
+            owl:someValuesFrom :Software-definedRadio ] ;
     :attack-id "EXF-0006.01" ;
     :definition "Threat actors may target software defined radios due to their software nature to setup exfiltration channels. Since SDRs are programmable, when combined with supply chain or development environment attacks, SDRs provide a pathway to setup covert exfiltration channels for a threat actor." ;
```
hack-sentinel: Change first sentence to "Threat actors may target **software-defined** radios due to their **software-based** nature to **set up** exfiltration channels."

hack-sentinel (changes requested):
Minor tweaks/adds and make sure to restore D3-EAL as class and id.

PR #482, opened 2025-12-24
T1489 restrictions

aamedina:
add restrictions to T1489 illustrating intended semantics of prescribes

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -786,6 +786,13 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :may-be-associated-with ;
     owl:inverseOf :may-modify .
 
+:may-be-prescribed-by a owl:ObjectProperty ;
+    rdfs:label "may-be-prescribed-by" ;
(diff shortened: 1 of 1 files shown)


aamedina:
re-run the build, looks like a transient HTTP error when CI downloads Jena

src/ontology/d3fend-protege.ttl
```diff
@@ -34046,7 +34077,22 @@ When system firmware verification fails a set of predefined responses is typical
...
+        [ a owl:Restriction ;
+            owl:onProperty :may-invoke ;
+            owl:someValuesFrom :TerminateProcess ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-prescribe ;
+            owl:someValuesFrom :ProcessTerminationEvent ],
+        [ a owl:Restriction ;
+            owl:onProperty :prescribes ;
+            owl:someValuesFrom [ a owl:Class ;
+                    owl:unionOf (
```
netfl0: Good idea but we need to figure out how to make this work with rest of the application before we merge this.

aamedina:
I'm going to split into two PRs, 1 just for prescribes alone and 1 to change this to just disables service application

PR #483, opened 2025-12-26
add restrictions for T1562.008

aamedina:


aamedina:
i feel like our apache died again

aamedina:
yet i can still download https://archive.apache.org/dist/jena/binaries/apache-jena-5.6.0.tar.gz

PR #490, opened 2025-12-26
T1621 abuses MFA services to produce auth requests for users

aamedina:


netfl0:
Feel like we need better artifacts for this one.

PR #504, opened 2026-01-27
D4S: Hardware Cryptographic Modules

ikiril01:
- Added new `CryptographicModule` class (sco of `DigitalInfoBearer`)
- Added Hardware/Software crypto module subclasses under `CryptographicModule`, including some specific variants like TPM.
- Added corresponding restrictions to existing classes (e.g., `LaptopComputer --> may-contain --> TrustedPlatformModule`).
- Added new CM (with refs): `CryptographicModuleOperatingModeRestriction`
- Added Crypto Module-related events


src/ontology/d3fend-protege.ttl
```diff
@@ -266,6 +266,12 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :associated-with ;
     :definition "x decodes y: Entity x transforms data y to a different form, usually through decompression." .
 
+:decrypts a owl:ObjectProperty ;
+    rdfs:label "decrypts" ;
+    rdfs:subPropertyOf :associated-with ;
+    owl:inverseOf :encrypts ;
```
aamedina: decrypts is not an inverse of encrypts

x decrypts y does not imply y encrypts x

src/ontology/d3fend-protege.ttl
```diff
@@ -19366,7 +19428,10 @@ LAN Access Mediation is a network security approach that manages and controls ac
     rdfs:label "Laptop Computer" ;
     skos:altLabel "Laptop",
         "Notebook" ;
-    rdfs:subClassOf :PersonalComputer ;
+    rdfs:subClassOf :PersonalComputer,
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :TrustedPlatformModule ] ;
```
aamedina: if you want to use may-contain, should this just be moved up in the computer hierarchy? any PC at least

src/ontology/d3fend-protege.ttl
```diff
@@ -6934,6 +6940,29 @@ Credential Scrubbing involves identifying and eliminating hard-coded credentials
...
+    rdfs:label "Cryptographic Module" ;
+    rdfs:subClassOf :DigitalInformationBearer,
+        [ a owl:Restriction ;
+            owl:onProperty :decrypts ;
+            owl:someValuesFrom :DigitalArtifact ],
+        [ a owl:Restriction ;
+            owl:onProperty :encrypts ;
+            owl:someValuesFrom :DigitalArtifact ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-produce ;
```
aamedina: why may produce

src/ontology/d3fend-protege.ttl
```diff
@@ -6934,6 +6940,29 @@ Credential Scrubbing involves identifying and eliminating hard-coded credentials
...
+
+:CryptographicModule a owl:Class ;
+    rdfs:label "Cryptographic Module" ;
+    rdfs:subClassOf :DigitalInformationBearer,
+        [ a owl:Restriction ;
+            owl:onProperty :decrypts ;
+            owl:someValuesFrom :DigitalArtifact ],
+        [ a owl:Restriction ;
+            owl:onProperty :encrypts ;
+            owl:someValuesFrom :DigitalArtifact ],
```
aamedina: this might be a candidate for owl:allValuesFrom, cryptographic modules encrypt/decrypt ONLY digital artifacts, not "at least one". also might make it add less pollution to the frontend

src/ontology/d3fend-protege.ttl
```diff
@@ -27094,6 +27165,9 @@ Yu, L., Zhang, W., Wang, J., & Yu, Y. (2017). SeqGAN: Sequence Generative Advers
         [ a owl:Restriction ;
             owl:onProperty :manages ;
             owl:someValuesFrom :ServiceApplicationProcess ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
+            owl:someValuesFrom :TrustedPlatformModule ],
```
aamedina, line 27846: maybe any host may-contain a TPM?

aamedina (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: Ah good point. I'll remove that.

src/ontology/d3fend-protege.ttl
ikiril01: True, we could do that. The other option is to use "contains" for the PCs that almost always contain TPMs these days - desktops, laptops, and servers.

ikiril01:
In scope for D4S v1.0.

PR #508, opened 2026-02-04
Object Property Definition Cleanup

ikiril01:
Went through all existing object property definitions and fixed:

1. Typos
2. Grammar issues
3. In a few cases, there were textual changes to the definition to make it clearer


aamedina (approved):
nice

PR #512, opened 2026-02-10
MRAM/ECCMemory/EEPROM

ikiril01:
- Added classes for 2 RAM subtypes (`ECCMemory`, `MRAM`) as well as EEPROM (sco of `SecondaryStorage`)


ikiril01:
This was added primarily because ECC Memory and EEPROM's are heavily used in space systems. Error correcting memory could be considered an implementation of EDAC https://en.wikipedia.org/wiki/Error_detection_and_correction. MRAM is similar in that because it stores data using magnetic states vs electric charges, it's highly radiation tolerant along with being non-volatile.

PR #515, opened 2026-02-25
EX-0007/Rad Hardening

ikiril01:
* Added new `ElectromagneticHardening` CM under Platform Hardening
* Added `RadiationHardening` Countermeasure + references
* Moved `RFShielding` under `ElectromagneticHardening`
* Added new `corrupts/may-corrupt` object property
* Added corresponding restrictions to EX-0007

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -819,7 +819,7 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
 :may-corrupt a owl:ObjectProperty ;
     rdfs:label "may-corrupt" ;
     rdfs:subPropertyOf :may-be-associated-with ;
-    :definition "x may-corrupt y: They entity x may corrupt y's information bearing content; that is, 'x corrupts y' may be true." .
+    :definition "x may-corrupt y: They entity x may corrupt y's information content; that is, 'x corrupts y' may be true." .
 
 :may-counter a owl:ObjectProperty ;
     rdfs:label "may-counter" ;
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -25460,7 +25460,10 @@ There are three core radiation hardening methodologies:
 
 :RAM a owl:Class ;
     rdfs:label "RAM" ;
-    rdfs:subClassOf :PrimaryStorage ;
+    rdfs:subClassOf :PrimaryStorage,
+        [ a owl:Restriction ;
+            owl:onProperty :carries ;
+            owl:someValuesFrom :MemoryExtent ] ; 
     rdfs:isDefinedBy <https://dbpedia.org/page/Random-access_memory> ;
     :definition "Random-access memory (RAM) is a form of computer memory that can be read and changed in any order, typically used to store working data and machine code." ;
     :synonym "Random-access Memory" .
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -14919,6 +14919,27 @@ Analyzing the interaction of a piece of code with a system while the code is bei
     rdfs:isDefinedBy <http://dbpedia.org/resource/Signal> ;
     :definition "Time-varying voltage or current that carries information." .
 
+:ElectromagneticHardening a :ElectromagneticHardening,
+        owl:Class,
+        owl:NamedIndividual ;
+    rdfs:label "Electromagnetic Hardening" ;
+    rdfs:subClassOf :PlatformHardening ;
+    :d3fend-id "D3-EMH" ;
+    :definition "The application of physical and material-level design measures to electronic systems, components, or facilities to reduce their susceptibility to damage or disruption from electromagnetic threats." ;
+    :kb-article """## How it works
+EM hardening operates on the principle of controlling the coupling path between an electromagnetic threat and the sensitive electronics it could affect. At the most fundamental level, this involves creating barriers that reflect, absorb, or redirect unwanted electromagnetic energy before it can induce damaging or disruptive currents in protected circuitry. The physical mechanisms exploited include the Faraday cage effect (conductive enclosures that attenuate external fields), skin-depth shielding (where conductive materials dissipate high-frequency fields before they penetrate), and transient suppression components (such as surge protectors and ferrite chokes) that clamp induced voltages at I/O interfaces.
+
+For threats at higher energy levels or involving ionizing radiation, such as nuclear EMP (NEMP) or space radiation, hardening extends beyond shielding to encompass radiation-tolerant component selection, redundant circuit architectures, and layout practices that minimize antenna-like structures susceptible to field coupling. The approach is inherently defense-in-depth: no single measure provides complete protection, so hardened systems typically layer multiple techniques across the facility, chassis, board, and component levels.
+
+## Considerations
+* Threat scope must be defined early: design choices differ significantly between defending against ambient RFI, conducted EMI on power lines, intentional jamming, HEMP (High-Altitude EMP), or ionizing radiation in space or nuclear environments.
+* Hardening can conflict with thermal management: fully sealed enclosures that maximize shielding often restrict airflow, requiring careful thermal design trade-offs.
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -25382,6 +25404,32 @@ Q-learning. Wikipedia. [Link](https://en.wikipedia.org/wiki/Q-learning).""" .
...
+            owl:someValuesFrom :PrimaryStorage ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-harden ;
+            owl:someValuesFrom :Processor ] ;
+    :d3fend-id "D3-RH" ;
+    :definition "Radiation hardening is the process of making electronic components and circuits resistant to damage or malfunction caused by high levels of ionizing radiation." ;
+    :kb-article """## How it works
+    There are three core radiation hardening methodologies:
+
+    1. Radiation Hardening by Process (RHBP): modifying the physical fabrication of a semiconductor (e.g., using SOI - Silicon on Insulator), offering the highest intrinsic protection. Usually the most expensive option as it requires a specialized semiconducter fabrication plant.
```
ikiril01: Should these be broken up into sub-techniques?

src/ontology/d3fend-protege.ttl
```diff
@@ -210,6 +210,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :associated-with ;
     :definition "x copy-of y: The subject x is a duplicate of the object y." .
 
+:corrupts a owl:ObjectProperty ;
```
aamedina, line 213: I feel like it's already too hard to choose between properties in D3FEND, and adding more subtly nuanced words is just going to compound the problem. We need a total refactor of object properties and definitive documentation for how to use them.

src/ontology/d3fend-protege.ttl
```diff
@@ -14909,6 +14919,27 @@ Analyzing the interaction of a piece of code with a system while the code is bei
...
 
+:ElectromagneticHardening a :ElectromagneticHardening,
+        owl:Class,
+        owl:NamedIndividual ;
+    rdfs:label "Electromagnetic Hardening" ;
+    rdfs:subClassOf :PlatformHardening ;
+    :d3fend-id "D3-EMH" ;
+    :definition "The application of physical and material-level design measures to electronic systems, components, or facilities to reduce their susceptibility to damage or disruption from electromagnetic threats." ;
+    :kb-article """## How it works
+EM hardening operates on the principle of controlling the coupling path between an electromagnetic threat and the sensitive electronics it could affect. At the most fundamental level, this involves creating barriers that reflect, absorb, or redirect unwanted electromagnetic energy before it can induce damaging or disruptive currents in protected circuitry. The physical mechanisms exploited include the Faraday cage effect (conductive enclosures that attenuate external fields), skin-depth shielding (where conductive materials dissipate high-frequency fields before they penetrate), and transient suppression components (such as surge protectors and ferrite chokes) that clamp induced voltages at I/O interfaces.
```
aamedina, line 15058: probably need an EM expert to review this kb article

src/ontology/d3fend-protege.ttl
```diff
@@ -15419,7 +15450,19 @@ Pennsylvania State University. (n.d.). Statistical Inference and Estimation. [Li
...
+            owl:someValuesFrom :Processor ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :ProcessorCacheMemory ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :ProcessorRegister ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :RAM ] ;
```
aamedina: impairs MemoryWord? MemoryExtent? Probably not just RAM.

src/ontology/d3fend-protege.ttl
```diff
@@ -25382,6 +25425,36 @@ Q-learning. Wikipedia. [Link](https://en.wikipedia.org/wiki/Q-learning).""" .
...
+        owl:NamedIndividual,
+        :RadiationHardening ;
+    rdfs:label "Radiation Hardening" ;
+    rdfs:subClassOf :ElectromagneticHardening,
+        [ a owl:Restriction ;
+            owl:onProperty :may-harden ;
+            owl:someValuesFrom :CentralProcessingUnit ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-harden ;
+            owl:someValuesFrom :EEPROM ],
```
aamedina: too specific

src/ontology/d3fend-protege.ttl
```diff
@@ -25382,6 +25425,36 @@ Q-learning. Wikipedia. [Link](https://en.wikipedia.org/wiki/Q-learning).""" .
...
+        [ a owl:Restriction ;
+            owl:onProperty :may-harden ;
+            owl:someValuesFrom :FlashMemory ],
```
aamedina: i don't think the mays make sense here

src/ontology/d3fend-protege.ttl
```diff
@@ -26576,7 +26649,7 @@ Queries for reverse resolution requests (that is, requests where IP(s) are sent
         owl:NamedIndividual,
         :RFShielding ;
     rdfs:label "RF Shielding" ;
-    rdfs:subClassOf :PlatformHardening ;
+    rdfs:subClassOf :ElectromagneticHardening ;
```
aamedina: aha got it

aamedina (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: That's fair, but we need some way of saying "attack x corrupts victim memory/device y". I couldn't think of a better way than adding an object property.

src/ontology/d3fend-protege.ttl
ikiril01: Good call.

src/ontology/d3fend-protege.ttl
ikiril01: Yeah, @hack-sentinel and I discussed this one, and the problem is that we can't easily link to an existing, more abstract class due to some prior modeling decisions. E.g., `may-harden :SecondaryStorage` would be ideal, but `SecondaryStorage` includes sub-classes like `CloudStorage` and `TertiaryStorage`, which would never be rad-hardened.

netfl0:
waiting on changes @ikiril01 correct?

ikiril01:
> waiting on changes @ikiril01 correct?

Correct, I need to make a few small changes.

ikiril01:
@netfl0 I've addressed the comments from my end, so this is ready for review again.

src/ontology/d3fend-protege.ttl
```diff
@@ -806,6 +811,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :may-be-associated-with ;
     :definition "to potentially have as contents or constituent parts; comprise; include." .
 
+:may-corrupt a owl:ObjectProperty ;
+    rdfs:label "may-corrupt" ;
+    rdfs:subPropertyOf :may-be-associated-with ;
+    :definition "x may-corrupt y: They entity x may corrupt y's information bearing content; that is, 'x corrupts y' may be true." .
```
netfl0: information bearing content;

should be

information content;

src/ontology/d3fend-protege.ttl
```diff
@@ -15419,7 +15450,19 @@ Pennsylvania State University. (n.d.). Statistical Inference and Estimation. [Li
...
+    rdfs:subClassOf :SPARTAExecutionTechnique,
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :CacheMemory ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :FlashMemory ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-corrupt ;
+            owl:someValuesFrom :MemoryExtent ],
```
netfl0, line 15590: this should probably reference Primary and Secondary storage versus narrow specification.

PR #516, opened 2026-03-02
D4S: Remote Command Authentication

ikiril01:
- Updated `RemoteCommand` to be a sub-class of `DigitalMessage` 
- Added Remote Command Authentication CM (for remote commands/telecommands). 
- Added SPARTA mappings around TCs


ikiril01:
In scope for D4S v1.0.

PR #519, opened 2026-03-09
D4S: SPARTA DE-0010/TBD

ikiril01:
* Added `FlightSoftwareEventLog` , sco of EventLog
* Added corresponding SPARTA restriction


src/ontology/d3fend-protege.ttl
```diff
@@ -19959,6 +19980,28 @@ Newer system call.""" .
     :definition "A record of events in the order of their occurrence." ;
     rdfs:seeAlso <http://dbpedia.org/resource/Chronology> .
 
+:LogAnalysis a :LogAnalysis,
```
netfl0: We model what is in the logs, not the logs themselves unless they are the direct target of attack. Drop this technique.

ikiril01: Got it.

ikiril01:
In scope for D4S v1.0.

ikiril01:
Need to determine if there are any reasonable countermeasures we can add against log flooding/log exhaustion attacks (which is what `DE-0010` is).

PR #528, opened 2026-03-31
REC-0005.04

ikiril01:
* Added mapping to REC-0005.04


ikiril01:
In scope for D4S v1.0.

ikiril01:
Closing this, as it has been merged into #558

PR #536, opened 2026-04-03
Clock/Timer Event Cleanup

ikiril01:
* Simplified Timer event taxonomy
* Fixed Clock/Timer event inheritance issues by removing top-level restrictions and adding leaf restrictions as needed

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -17940,17 +17940,12 @@ Hardware process isolation is commonly implemented through Direct Memory Access 
     :definition "A device driver for a hardware clock." ;
     rdfs:seeAlso <https://docs.kernel.org/admin-guide/rtc.html> .
 
-:HardwareClockEvent a owl:Class ;
-    rdfs:label "Hardware Clock Event" ;
+:HardwareClockUpdateEvent a owl:Class ;
+    rdfs:label "Hardware Clock Update Event" ;
     rdfs:subClassOf :ClockEvent,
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -1758,138 +1758,6 @@ skos:prefLabel a owl:AnnotationProperty .
 
 ### Classes
 
-:Transponder a owl:Class ;
```
ikiril01: I blame ttlfmt for this.

src/ontology/d3fend-protege.ttl
```diff
@@ -40083,6 +40049,14 @@ Temporal difference learning. Wikipedia.  [Link](https://en.wikipedia.org/wiki/T
     rdfs:isDefinedBy <https://dbpedia.org/resource/Timer> ;
     :definition "A timer or countdown timer is a type of clock that starts from a specified time duration and stops upon reaching 00:00. It can also usually be stopped manually before the whole duration has elapsed. An example of a simple timer is an hourglass. Commonly, a timer triggers an alarm when it ends. A timer can be implemented through hardware or software." .
 
+:TimerConfigurationEvent a owl:Class ;
+    rdfs:label "Timer Configuration Event" ;
+    rdfs:subClassOf :TimerEvent,
+        [ a owl:Restriction ;
+            owl:onProperty :has-participant ;
```
MTolsma01, line 40056: This should probably be modifies rather than has-participant

PR #545, opened 2026-04-14
ATT&CK for ICS mappings for T0839-T0845

holtorama:


src/ontology/d3fend-protege.ttl
```diff
@@ -29808,14 +29820,26 @@ When system firmware verification fails a set of predefined responses is typical
 :T0842 a owl:Class ;
     rdfs:label "Network Sniffing - ATTACK ICS" ;
     skos:prefLabel "Network Sniffing" ;
-    rdfs:subClassOf :ATTACKICSDiscoveryTechnique ;
+    rdfs:subClassOf :ATTACKICSDiscoveryTechnique,
+        [ a owl:Restriction ;
```
aamedina, line 29824: this might be more of a general assumption in enterprise contexts but less likely in OT

aamedina (changes requested):

PR #558, opened 2026-04-22
D4S: Signal Hardening CMs

ikiril01:
* Added new top-level Signal Hardening CM
* Added Spectral Manipulation (Spread Spectrum) as an sco of Signal Hardening
* Added Digital Beamforming as an sco of Signal Hardening


src/ontology/d3fend-protege.ttl
```diff
@@ -28517,6 +28543,34 @@ The goal is for homophones to be encoded to the same representation so that they
...
+
+## Considerations
+* Processing gain is the central design trade-off: greater spread yields stronger jamming resistance and lower probability of intercept, but demands wider spectrum access and more complex hardware.
+* Synchronization is a critical vulnerability: the receiver must maintain precise alignment with the transmitter's spreading code; loss of synchronization collapses the link entirely.
+* The spreading code must remain secret and unpredictable: a compromised or predictable code eliminates most of the defensive benefit and can expose the signal to targeted jamming or spoofing.
+* Spectrum availability limits the achievable spread: regulatory, physical, or operational constraints on usable bandwidth cap the maximum processing gain in any given deployment environment.
+* Spectral manipulation is not encryption: it obscures the existence and location of a signal, but does not by itself protect its contents.
```
MTolsma01: Suggested re-wording: Spectral manipulation is not encryption: it fragments the message carried on a signal across the spectrum to make interception more difficult, but does not by itself protect the contents.

ikiril01: Thanks - I like the re-wording, so I'll use that.

src/ontology/d3fend-protege.ttl
```diff
@@ -121,6 +121,11 @@
     rdfs:subPropertyOf :semantic-relation ;
     :definition "x broader-transitive y: The entity x represents a more general concept than entity y, including indirect or hierarchical relationships where x encompasses y through intermediate entities." .
 
+:captures a owl:ObjectProperty ;
+    :definition "x captures y: The entity x acquires, records, or stores a copy of communications data y from any source or state (e.g., in transit, at rest, or from storage) such that x can subsequently access, analyze, or use y." ;
+    :label "captures" ;
+    :subPropertyOf :associated-with .
```
netfl0, line 128: going to need domain and range

src/ontology/d3fend-protege.ttl
```diff
@@ -630,6 +635,11 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:subPropertyOf :associated-with ;
     :definition "x instructs y: A subject x delivers machine instructions to object y." .
 
+:intercepts a owl:ObjectProperty ;
+    :definition "x intercepts y: The entity x captures communications data y while y is in transit between other communicating parties, where x is not an intended endpoint of that communication." ;
+    :label "intercepts" ;
+    :subPropertyOf :captures .
```
netfl0, line 643: going to need domain and range

src/ontology/d3fend-protege.ttl
```diff
@@ -27992,6 +28057,18 @@ Another difference are the capabilities of the mechanism:
     rdfs:isDefinedBy <http://dbpedia.org/resource/Signal> ;
     :definition "In electronics and telecommunications, signal refers to any time-varying voltage, current, or electromagnetic wave that carries information." .
 
+:SignalHardening a owl:Class,
```
netfl0, line 28138: I think the more established term is transsec ala https://en.wikipedia.org/wiki/Transmission_security. Need to hash this out.

src/ontology/d3fend-protege.ttl
```diff
@@ -46621,4 +46748,4 @@ With keystroke dynamics the biometric template used to identify an individual is
 
 <wptmp:entity#Reference%20-%20DNS%20Whitelist%20(DNSWL)%20Email%20Authentication%20Method%20Extension> :kb-author "Alessandro Vesely" .
 
-### Serialized using the ttlser deterministic serializer v1.2.1
```
netfl0: This should not be happening, but it could be my fault. Please hunt this down for me.

netfl0 (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: Agreed. One option is to add "TRANSEC" as a top-level CM and then move SignalHardening et. al under it.

PR #564, opened 2026-05-12
Fixed Reference-CAR-2021-01-007

ikiril01:
Removed extraneous "d" in :kb-mitre-analysis for `:Reference-CAR-2021-01-007%3ADetectingTamperingOfWindowsDefenderCommandPrompt_MITRE`.


netfl0:
Remove the property altogether, good find.

ikiril01:
> Remove the property altogether, good find.

Sounds good, I just removed it altogether.

PR #566, opened 2026-05-13
update to ATT&CK v19

### LINKED ISSUE #565: Update to ATT&CK v19
May be some complications with Defense Evasion split

aamedina:
* Defense Evasion technique deprecated; split into Stealth and Defense Impairment Technique

closes #565


src/ontology/d3fend-protege.ttl
```diff
@@ -6825,6 +6819,13 @@ CFI is most effective when combined with complementary defenses such as stack ca
     rdfs:subClassOf :ControlFlowGraph ;
     :definition "A control flow policy is a subset of the possible control flow transfers computed from a program's control flow graph. It defines only the expected and allowed control flow transfers and is enforced by control flow integrity." .
 
+:ControlServer a owl:Class ;
```
netfl0, line 6822: i was not expecting to see new artifacts in this. Lets keep the mapping and update separate. Please create another mappings PR which will happen AFTER we merge in the new v19 content.

aamedina: this isn't a new artifact it just printed in a different order

PR #589, opened 2026-06-17
D4S: Ground Station Artifacts + Telemetry

ikiril01:
- Added a few ground station artifacts + SPARTA mappings

This is for D4S 1.0.


src/ontology/d3fend-protege.ttl
```diff
@@ -39978,6 +40032,15 @@ Defense Impairment consists of techniques that degrade, disable, or undermine th
     :definition "A telecommand or telecontrol is a command sent to control a remote system or systems not directly connected (e.g. via wires) to the place from which the telecommand is sent." ;
     :synonym "Remote Control Command" .
 
+:Telemetry a owl:Class ;
```
MTolsma01, line 40468: I'm wondering if we should create telemetry class and a subclass of space vehicle telemetry. While your definition covers both, but I think tying the two so closely together will create confusion down the line.

ikiril01: Good call, I refactored Telemetry and made a SpacecraftTelemetry sub-class.

PR #602, opened 2026-07-02
D4S: PNT Modeling (incl. GNSS)

ikiril01:
- Added more specific GNSS artifacts
- Added GNSSNavigationMessageAuthentication CM
- Added corresponding SPARTA mappings


ikiril01:
Also: add `may-contain Firmware` restrictions to Receiver/Transmitter/Transceiver.

netfl0:
move toward generalize PNT terminology versus GNSS/GPS

PR #608, opened 2026-07-15
Add D3FEND artifact restrictions for ATLAS AML.T0088

flying-coyote:
Maps `AML.T0088` (Generate Deepfakes) to existing D3FEND artifacts:

- `:produces :DigitalMedia` (Digital Media) — *"create synthetic media (i.e. imagery, video, audio, and text)"*

Same restriction shape as `:T1003.001`; quotes are verbatim from the technique's `d3f:definition`.

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -3802,7 +3802,10 @@ Adversaries may use the gathered victim data to Create Deepfakes and impersonate
 :AML.T0088 a owl:Class ;
     rdfs:label "Generate Deepfakes - ATLAS" ;
     skos:prefLabel "Generate Deepfakes" ;
-    rdfs:subClassOf :ATLASAIAttackStagingTechnique ;
+    rdfs:subClassOf :ATLASAIAttackStagingTechnique,
+        [ a owl:Restriction ;
+            owl:onProperty :produces ;
+            owl:someValuesFrom :DigitalMedia ] ;
     :attack-id "AML.T0088" ;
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -3643,7 +3643,16 @@ Adversaries may use the gathered victim data to Create Deepfakes and impersonate
 :AML.T0088 a owl:Class ;
     rdfs:label "Generate Deepfakes - ATLAS" ;
     skos:prefLabel "Generate Deepfakes" ;
-    rdfs:subClassOf :ATLASAIAttackStagingTechnique ;
+    rdfs:subClassOf :ATLASAIAttackStagingTechnique,
+        [ a owl:Restriction ;
+            owl:onProperty :produces ;
+            owl:someValuesFrom :DigitalAudio ],
```
aamedina: might just be worth going with one restriction here to d3f:DigitalMedia to account for any kind of synethetic media

aamedina (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: Agreed, it also means you can use `:produces` here, whereas `:may-produce` would need to be used with multiple restrictions.

flying-coyote:
Consolidation is in: single `:produces :DigitalMedia` restriction per your review, using `:produces` rather than `:may-produce` per @ikiril01. Rebased onto current develop; quote re-verified against v2026.06. Re-requesting review.

PR #621, opened 2026-07-17
Updated SPARTA Mappings

ikiril01:
- Updated some existing SPARTA mappings for accuracy (different artifacts/relationships)
- Deprecated a few mappings for overly abstract techniques (e.g., side channel attacks)

### DIFF
--- a/src/ontology/d3fend-protege.ttl
+++ b/src/ontology/d3fend-protege.ttl
@@ -14004,7 +14004,7 @@ Architectural limitations in ground station software can also enable deception w
     rdfs:subClassOf :DE-0003,
         [ a owl:Restriction ;
             owl:onProperty :modifies ;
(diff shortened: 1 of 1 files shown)


src/ontology/d3fend-protege.ttl
```diff
@@ -16117,7 +16126,7 @@ Pennsylvania State University. (n.d.). Statistical Inference and Estimation. [Li
     rdfs:subClassOf :EX-0005,
         [ a owl:Restriction ;
             owl:onProperty :produces ;
-            owl:someValuesFrom :OTModifyDeviceConfigurationCommand ] ;
+            owl:someValuesFrom :BusMessage ] ;
```
netfl0, line 16129: produces OT Protocol Message

ikiril01: Done

PR #623, opened 2026-07-20
Network Interfaces

ikiril01:
# Add logical network interface artifact hierarchy

## Summary

This change adds a logical `NetworkInterface` artifact under `LocalResource` and 14 representative subclasses covering:

- physical versus virtual implementation;
- Ethernet, wireless LAN, bluetooth, cellular, bus, and CAN interfaces;
- loopback, point-to-point, tunnel, bridge, VLAN, and aggregated interfaces.

It leaves the existing `NetworkInterfaceCard` class unchanged as the physical hardware component. `HardwareBackedNetworkInterface` uses an `implemented-by some NetworkInterfaceCard` restriction to connect the logical and physical layers without conflating them.

## Motivation

D3FEND can represent a physical NIC, a network, a link, a host/node, frames, messages, and traffic, but it lacks the logical endpoint/resource exposed to local software. Consequently, common system objects such as `eth0`, `wlan0`, `lo`, `br0`, `tun0`, and `can0` cannot currently be typed precisely.

The distinction is security-relevant: a process can be authorized to use, monitor, configure, or abuse a logical interface independently of the physical controller and independently of the network to which the interface is attached.

## Added classes

- `NetworkInterface`
- `HardwareBackedNetworkInterface`
- `VirtualNetworkInterface`
- `EthernetNetworkInterface`
- `WirelessNetworkInterface`
- `WirelessLANNetworkInterface`
- `CellularNetworkInterface`
- `BusNetworkInterface`
- `CANNetworkInterface`
- `LoopbackNetworkInterface`
- `NetworkTunnelInterface`
- `NetworkBridgeInterface`
- `VLANNetworkInterface`
- `AggregatedNetworkInterface`

## Modeling notes

The hierarchy is intentionally non-disjoint and supports multiple typing. Examples:

- a physical `can0`: `HardwareBackedNetworkInterface` + `CANNetworkInterface`;
- a virtual `vcan0`: `VirtualNetworkInterface` + `CANNetworkInterface`;
- a VLAN subinterface: `VLANNetworkInterface`, which entails both `VirtualNetworkInterface` and `EthernetNetworkInterface`;
- a TUN device: `NetworkTunnelInterface` + `PointToPointNetworkInterface`.

Host-specific interface names remain instance labels and are not added as classes.

## Scope

This PR changes only `src/ontology/d3fend-protege.ttl`. It does not:

- reclassify `NetworkInterfaceCard`;
- add operating-system-specific individuals;
- add interface names, addresses, administrative state, or namespace properties;
- model sockets, links, networks, traffic, messages, or frames as interfaces.


src/ontology/d3fend-protege.ttl
```diff
@@ -25521,6 +25676,16 @@ A known potential limitation of PACs concerns signing gadgets. Under certain cir
     :kb-article """## References
 Pennsylvania State University. (n.d.). Statistical Inference and Estimation. [Link](https://online.stat.psu.edu/stat504/lesson/statistical-inference-and-estimation)""" .
 
+:PointToPointNetworkInterface a owl:Class ;
+    rdfs:label "Point-to-Point Network Interface" ;
+    rdfs:subClassOf :NetworkInterface,
+        [ a owl:Restriction ;
+            owl:onProperty :connected-to ;
+            owl:someValuesFrom :LogicalLink ] ;
```
aamedina, line 25690: not sure I get this one

aamedina (commented):
really like this overall, we can all debate specific restrictions later

ikiril01:
> really like this overall, we can all debate specific restrictions later

Appreciate the review!

src/ontology/d3fend-protege.ttl
ikiril01: Yeah I'll remove this.

src/ontology/d3fend-protege.ttl
```diff
@@ -22062,17 +22173,51 @@ Network Access Mediation is a crucial process in telecommunications and IT netwo
...
+    rdfs:subClassOf :LocalResource,
+        [ a owl:Restriction ;
+            owl:onProperty :connected-to ;
+            owl:someValuesFrom :Network ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-be-contained-by ;
+            owl:someValuesFrom :ComputerNetworkNode ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-be-contained-by ;
+            owl:someValuesFrom :OperatingSystem ],
```
ioggstream, line 22199: While this might be useful on the representation side, this could be :managed-by

ikiril01: Good point! `managed-by` works better, I'll make the change.

ikiril01: Hmm I guess we don't actually have `managed-by`, but I'll add it as an inverse of `manages`.

src/ontology/d3fend-protege.ttl
```diff
@@ -710,6 +710,12 @@ Moving forward different distinctions of kinds of has-part (contains) relationsh
     rdfs:isDefinedBy <http://wordnet-rdf.princeton.edu/id/02236692-v> ;
     :definition "x loads y: The technique or process x transfers a software from storage y to a computer's memory for subsequent execution." .
 
+:managed-by a owl:ObjectProperty ;
+    rdfs:label "managed-by" ;
```
ioggstream, line 714: 👍

PR #630, opened 2026-07-22
#164 subset - ot controller model artifact subset

holtorama:


src/ontology/d3fend-protege.ttl
```diff
@@ -23847,6 +23938,28 @@ Modbus: Exception: Gateway Target Device Failed to Respond""" ;
...
+    rdfs:label "OT Merging Unit" ;
+    rdfs:subClassOf :OTController,
+        [ a owl:Restriction ;
+            owl:onProperty :contains ;
+            owl:someValuesFrom :Clock ],
+        [ a owl:Restriction ;
+            owl:onProperty :contains ;
+            owl:someValuesFrom :OTPowerSupply ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-contain ;
```
ikiril01: Do these units actually contain the transformer, or are they simply connected to it? If the latter, I recommend changing this to `connected-to`.

PR #667, opened 2026-09-17
Added T1046 Mapping

ikiril01:
Added D3FEND mapping to `T1046`, using an existing D3FEND artifact.


aamedina (commented):
what about enumerates service application process?

ikiril01:
> what about enumerates service application process?

Good point, I think that's reasonable to add along with the `produces NetworkTraffic` restriction.

aamedina (changes requested):
weird typo got in the definition

ikiril01:
> weird typo got in the definition

Whoops, fixed now.

PR #671, opened 2026-09-18
Added T1216 Mapping

ikiril01:
Added D3FEND mapping to `T1216`, using an existing D3FEND artifact.


src/ontology/d3fend-protege.ttl
```diff
@@ -35024,7 +35015,10 @@ When system firmware verification fails a set of predefined responses is typical
 
 :T1216 a owl:Class ;
     rdfs:label "System Script Proxy Execution" ;
-    rdfs:subClassOf :StealthTechnique ;
+    rdfs:subClassOf :StealthTechnique,
+        [ a owl:Restriction ;
+            owl:onProperty :executes ;
```
netfl0: invokes

netfl0 (changes requested):

PR #673, opened 2026-09-18
Additional T1546.xx Mappings

ikiril01:
Added D3FEND mappings to several subclasses of `T1546` that weren't previously covered, using existing D3FEND artifacts.


src/ontology/d3fend-protege.ttl
```diff
@@ -36938,19 +36920,34 @@ When system firmware verification fails a set of predefined responses is typical
 
 :T1546.016 a owl:Class ;
     rdfs:label "Installer Packages" ;
-    rdfs:subClassOf :T1546 ;
+    rdfs:subClassOf :T1546,
+        [ a owl:Restriction ;
+            owl:onProperty :may-execute ;
```
netfl0: prop should be invoke

netfl0 (changes requested):

PR #679, opened 2026-09-21
Added T1652 Mapping

ikiril01:
Added D3FEND mapping to `T1652`, using an existing D3FEND artifact.


src/ontology/d3fend-protege.ttl
```diff
@@ -39695,7 +39677,22 @@ When system firmware verification fails a set of predefined responses is typical
...
+    rdfs:subClassOf :DiscoveryTechnique,
+        [ a owl:Restriction ;
+            owl:onProperty :enumerates ;
+            owl:someValuesFrom :HardwareDriver ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-access ;
+            owl:someValuesFrom :OperatingSystemFile ],
+        [ a owl:Restriction ;
+            owl:onProperty :may-create ;
+            owl:someValuesFrom :Process ],
```
netfl0: Why don't we add something for a kernel driver process, this will require more research about different driver instantiation approaches across the various kinds of kernels.

We may want to just remove this from this one, it will produce lots of matches.

netfl0 (changes requested):

src/ontology/d3fend-protege.ttl
ikiril01: Sounds good, I removed it.

PR #680, opened 2026-09-22
Added T1654 Mapping

ikiril01:
Added D3FEND mapping to `T1654`, using existing D3FEND artifacts.


netfl0 (changes requested):
Let's rather scope to EventLog.

ikiril01:
> Let's rather scope to EventLog.

So `:enumerates EventLog` then?

